PR #9, opened 2023-03-22
Cairo runner tests

DomHenderson:
This adds the initial framework for tests that run sierra both via mlir and via starkware's cairo runner and compare the results. Could be changed to use the ExecutionEngine, but at that point I think it's worth making a PR to have the cli's run function get the result from the execution engine and return it and use that which imo should be a separate PR. The wrappers generated here would also be useful for that as they could be repurposed to return a vector/array/tensor of felts rather than printing it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ Cargo.lock
 /target
 
 llvm/
+
+# Added so we can have our own specific development scripts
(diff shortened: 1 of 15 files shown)


edg-l (approved):
lgtm

PR #19, opened 2023-03-22
streamline variable storage

edg-l:
Reduces boilerplate when getting the value, hiding whether its from a block argument or a operation result.


DomHenderson (approved):
Looks good to me

PR #32, opened 2023-03-23
Dataflow handling

DomHenderson:
The plan

- Split processing into funcs->blocks->state
- Funcs processing consists of working out argument and return types for each function and storing them
- For block processing, define statement ranges for each block, as well as successors and predecessors. Then iterate through the statements in each block to determine the locally created and required variables. Then, use the stored list of predecesssors to propagate any unmet variable requirements. Once this has been done, the Region and Blocks can be created as we know what arguments they need, what types they are, and what variable indices they correspond to (which needs saving in storage)
- Statement processing is now relatively simple as statements can be processed one block at a time, in any order of blocks


edg-l (approved):
lgtm awesome work!

igaray (approved):

PR #37, opened 2023-03-27
Improve `felt252` addition and subtraction modulo.

azteca1998:


sierra2mlir/src/compiler.rs
```diff
@@ -1031,3 +1038,22 @@ impl<'ctx> Compiler<'ctx> {
         }
     }
 }
+
+// TODO: Add other supported comparisons.
+//   Source: https://mlir.llvm.org/docs/Dialects/ArithOps/#arithcmpi-mlirarithcmpiop
+#[derive(Debug)]
```
edg-l, line 1044: I always like simple enums like this to impl Copy

edg-l (dismissed):
lgtm

edg-l (approved):

Issue #26, opened 2023-03-22
Libfuncs: enm


igaray:
[Sierra implementation](https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra/src/extensions/modules/enm.rs)
- [x] enum_init
- [x] enum_match

The `enum_snapshot_match` libfunc is covered by #93

edg-l:
https://mapping-high-level-constructs-to-llvm-ir.readthedocs.io/en/latest/basic-constructs/unions.html#tagged-unions

PR #43, opened 2023-03-28
enum type and enum_init

edg-l:
Using this as reference: https://mapping-high-level-constructs-to-llvm-ir.readthedocs.io/en/latest/basic-constructs/unions.html#tagged-unions

Basically holds a tag indicating the variant, the enum itself is then a llvm i8 array with enough bytes to hold the biggest type.

This is how most implementations of tagged unions are done i think (enums).

related to #26


sierra2mlir/src/types/mod.rs
```diff
@@ -93,6 +93,77 @@ impl<'ctx> Compiler<'ctx> {
...
+                        GenericArg::UserType(x) => x,
+                        _ => {
+                            unreachable!("first arg on struct libfunc should always be a user type")
+                        }
+                    };
+
+                    for gen_arg in type_decl.long_id.generic_args.iter().skip(1) {
+                        let gen_arg_type_id = match gen_arg {
+                            GenericArg::Type(gen_arg_typeid) => gen_arg_typeid.id.to_string(),
+                            GenericArg::UserType(user_type_id) => user_type_id.id.to_string(),
```
DomHenderson, line 114: Since the first argument is skipped, this match should only need an arm for `Type` and one that's marked unreachable

sierra2mlir/src/libfuncs/sierra_enum.rs
```diff
@@ -0,0 +1,106 @@
...
+        &'ctx self,
+        func_decl: &LibfuncDeclaration,
+        parent_block: BlockRef<'ctx>,
+        storage: Rc<RefCell<Storage<'ctx>>>,
+    ) -> Result<()> {
+        let id = Self::normalize_func_name(func_decl.id.debug_name.as_ref().unwrap().as_str())
+            .to_string();
+
+        let enum_arg_type = match &func_decl.long_id.generic_args[0] {
+            GenericArg::UserType(_) => todo!(),
```
DomHenderson, line 23: This should be unreachable like the other non-Type arms

DomHenderson (changes requested):
Just a couple nitpicks, otherwise this looks great

sierra2mlir/tests/example_compiles.rs
```diff
@@ -15,4 +15,15 @@ macro_rules! impl_tests {
...
+    casts,
+    destructure,
+    felt_is_zero,
+    fib,
+    fib_simple,
+    print_test,
+    program,
+    simple,
+    types,
+    simple_enum
```
azteca1998, line 28:
```suggestion
    casts,
    destructure,
    felt_is_zero,
    fib,
    fib_simple,
    print_test,
    program,
    simple,
    simple_enum,
    types,
```

sierra2mlir/src/types/mod.rs
```diff
@@ -93,6 +93,76 @@ impl<'ctx> Compiler<'ctx> {
...
+                        }
+                    }
+
+                    let bytes = biggest_type.get_width().unwrap() / 8;
+                    let enum_memory_array =
+                        Type::parse(&self.context, &format!("!llvm.array<{bytes} x i8>",))
+                            .expect("create the enum storage type.");
+
+                    // for now the tag is a u16 = 65535 variants.
+                    // TODO: make the tag size variable?
```
azteca1998, line 149: The enum's tag will be a nice use case for the extra 4 bits in a `felt252`.

sierra2mlir/src/types/mod.rs
```diff
@@ -93,6 +93,76 @@ impl<'ctx> Compiler<'ctx> {
...
+                    }
+
+                    let enum_variant_sierra_types = unique_enum_types
+                        .iter()
+                        .map(|x| storage.types.get(x).cloned().expect("type should exist"))
+                        .collect_vec();
+                    let unique_enum_variant_types =
+                        enum_variant_sierra_types.iter().map(SierraType::get_type).collect_vec();
+
+                    // The enum is a struct consiting of the tag and enough memory to allocate the biggest element.
```
azteca1998, line 133: Maybe it'd be nice to mention C unions, as it's basically the same.

sierra2mlir/src/types/mod.rs
```diff
@@ -93,6 +93,76 @@ impl<'ctx> Compiler<'ctx> {
...
+
+                    // in case the enum has variants with the same type, just store a uniqued one.
+                    let mut unique_enum_types = HashSet::new();
+                    // all the enum variant types, even repeated, for later use when storing the variant value.
+                    let mut enum_variant_types = vec![];
+
+                    let _user_type = match &type_decl.long_id.generic_args[0] {
+                        GenericArg::UserType(x) => x,
+                        _ => {
+                            unreachable!("first arg on struct libfunc should always be a user type")
```
azteca1998, line 107: I'd use `todo!()` if we're going to add an error type, or just `panic!()` directly because `unreachable!()` implies that the code shouldn't be reachable, but we can't really be sure about that since the sierra code is provided externally.

sierra2mlir/src/types/mod.rs
```diff
@@ -93,6 +93,76 @@ impl<'ctx> Compiler<'ctx> {
...
+                        }
+                    };
+
+                    for gen_arg in type_decl.long_id.generic_args.iter().skip(1) {
+                        let gen_arg_type_id = match gen_arg {
+                            GenericArg::Type(gen_arg_typeid) => gen_arg_typeid.id.to_string(),
+                            _ => unreachable!("should always be a type"),
```
azteca1998, line 114: Same as before.

azteca1998 (dismissed):
Awesome.

sierra2mlir/src/types/mod.rs
edg-l: i think the link in a comment already mentions it

sierra2mlir/src/types/mod.rs
edg-l: i think unreachable is fine here, since this is the safe version it is essentially a panic! wrapper with a semantic

azteca1998 (approved):

PR #49, opened 2023-03-28
Store more comprehensive information about implementation of libfuncs

DomHenderson:
# SierraLibFunc

## Description

We've been introducing special cases around quite a few libfuncs, this is (most of) an attempt to instead create a representation of the patterns we use, along the lines of SierraType. This should allow dataflow and statement processing to be simplified, as well as removing the need for a member of storage for every type of constant. Cases included in SierraLibFunc:

- libfuncs that are implemented as constant values
- normal functions
- libfuncs that are ignored entirely
- implementations that only use a subset of the arguments of the sierra func (e.g. bitwise)
- branching functions with differing returns (currently felt252_is_zero, but most aimed at enum_match)

I plan to improve legibility by replacing the enum with a trait implemented by multiple structs, but here is a proof of concept

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


edg-l (approved):
lgtm

PR #53, opened 2023-03-29
fix cli run

edg-l:
the dprintf symbol was not declared when running

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ compile-example-optimized: check-mlir
 	RUST_LOG="debug" cargo r -- compile examples/simple.sierra --optimize
 
 build-examples: check-mlir
-	cargo r -- compile examples/print_test.sierra -o examples/print_test.mlir -m -p 1
+	cargo r -- compile examples/print_test.sierra -o examples/print_test.mlir -m
 	cargo r -- compile examples/simple.sierra -o examples/simple.mlir
 	cargo r -- compile examples/simple.sierra -o examples/simple-optimized.mlir --optimize
 	cargo r -- compile examples/simple_enum.sierra -o examples/simple_enum.mlir
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -85,7 +85,7 @@ fn main() -> color_eyre::Result<()> {
         Commands::Compile { input, optimize, output, debug, main_print, print_target } => {
             let code = fs::read_to_string(input)?;
             let mlir_output =
-                sierra2mlir::compile(&code, optimize, debug, main_print.then_some(print_target))?;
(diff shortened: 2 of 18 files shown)


cli/src/main.rs
```diff
@@ -74,23 +82,20 @@ fn main() -> color_eyre::Result<()> {
...
 
     match args.command {
-        Commands::Compile { input, optimize, output, debug, main_print } => {
-            let main_print = main_print.or(Some(1)).filter(|x| *x >= 0);
-
+        Commands::Compile { input, optimize, output, debug, main_print, print_target } => {
             let code = fs::read_to_string(input)?;
-            let mlir_output = sierra2mlir::compile(&code, optimize, debug, main_print)?;
+            let mlir_output =
+                sierra2mlir::compile(&code, optimize, debug, main_print.then_some(print_target))?;
```
azteca1998, line 88: I think this is wrong. Disabling `main_print` shouldn't disable all the prints.

cli/src/main.rs
```diff
@@ -74,23 +82,20 @@ fn main() -> color_eyre::Result<()> {
...
                 println!("{mlir_output}");
             }
         }
-        Commands::Run { function, input, main_print } => {
-            let main_print = main_print.or(Some(1)).filter(|x| *x >= 0);
-
+        Commands::Run { function, input, main_print, print_target } => {
             let code = fs::read_to_string(input)?;
-            let engine = sierra2mlir::execute(&code, main_print)?;
+            let engine = sierra2mlir::execute(&code, main_print.then_some(print_target))?;
```
azteca1998, line 98: Same as before.

sierra2mlir/src/lib.rs
```diff
@@ -70,9 +69,19 @@ pub fn compile(
 
 pub fn execute(code: &str, main_print: Option<i32>) -> Result<ExecutionEngine, color_eyre::Report> {
     let mut compiler = Compiler::new(code, main_print)?;
+
+    if main_print.is_some() {
+        compiler.create_printf()?;
+    }
```
azteca1998, line 75:
```suggestion
    if <are prints enabled> {
        compiler.create_printf()?;
    }
```

Makefile
```diff
@@ -19,7 +19,7 @@ compile-example-optimized: check-mlir
 	RUST_LOG="debug" cargo r -- compile examples/simple.sierra --optimize
 
 build-examples: check-mlir
-	cargo r -- compile examples/print_test.sierra -o examples/print_test.mlir -m
+	cargo r -- compile examples/print_test.sierra -o examples/print_test.mlir -m -p 1
```
azteca1998, line 22: `-p 1` is the default, so this isn't really necessary.

azteca1998 (approved):

PR #56, opened 2023-03-29
add enum print, fix enum bug

edg-l:
adds enum print support, fixes issues with the gep that i was using wrongly before

this also adds the prints for u8,u16,u32,u64,u128

not sure about u128 working though


azteca1998 (approved):
Nice.

azteca1998 (approved):
Nice.

DomHenderson (changes requested):
Can we get some comparison tests on this now that it can print?

DomHenderson (approved):

PR #64, opened 2023-03-30
move control flow inline libfuncs to their own module, start enum_match

edg-l:
work in progress

this pr moves the libfuncs implemented inline (such as felt is zero and jump) into a module of its own for cleaner code

it also implements enum_match


sierra2mlir/src/compiler.rs
```diff
@@ -1065,9 +1090,11 @@ impl<'ctx> Compiler<'ctx> {
...
     ///
     /// a::a::name -> a_a_name()
-    pub fn normalize_func_name(name: &str) -> Cow<str> {
-        let re = Regex::new(r"[:-]+").unwrap();
-        re.replace_all(name, "_")
+    pub const fn normalize_func_name(name: &str) -> Cow<str> {
+        Cow::Borrowed(name)
+        // todo: remove as it is not necessary and causes conflicts with type names.
+        //let re = Regex::new(r"[:-]+").unwrap();
+        //re.replace_all(name, "_")
```
azteca1998, line 1097: Maybe this function can be removed completely?

sierra2mlir/src/utility.rs
```diff
@@ -199,55 +199,52 @@ impl<'ctx> Compiler<'ctx> {
...
 
         if let SierraType::Enum {
-            ty,
+            ty: _,
             tag_type,
             storage_bytes_len: _,
-            storage_type: _,
+            storage_type,
             variants_types,
         } = enum_type
```
azteca1998, line 212: When destructuring a struct-like construct without needing all its members you can avoid having to specify them all by doing something like this:
```rs
if let SierraType::Enum {
    tag_type,
    storage_type,
    variants_types,
    ..
} = enum_type
```

sierra2mlir/src/utility.rs
```diff
@@ -258,9 +255,20 @@ impl<'ctx> Compiler<'ctx> {
...
                 .collect_vec();
 
+            let enum_felt_width = enum_type.get_felt_representation_width();
+            println!("Enum felt width: {enum_felt_width}");
+
             for (i, (block, var_ty)) in blocks.iter().enumerate() {
+                let variant_felt_width = var_ty.get_felt_representation_width();
+                println!("variant_felt_width: {variant_felt_width}");
+                let unused_felt_count = enum_felt_width - 1 - variant_felt_width;
+                println!("unused_felt_count: {unused_felt_count}");
```
azteca1998, line 265: Are those `println!()` intended?

azteca1998 (approved):

PR #65, opened 2023-03-30
Rewrite `Makefile` to make it more compatible.

azteca1998:
# Rewrite `Makefile` to make it more compatible

## Description

The makefile now should work on:
  - Machines with `LLVM_SYS_160_PREFIX`
  - Machines with multiple LLVM versions.
  - Machines with a single LLVM of the right version.

Other changes:
  - The LLVM installation rule has been moved to its own script.
  - Remove compiled cairo stuff (`*.sierra`, `*.mlir`, `*.ll`).
  - Add rules to build the removed stuff automatically.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


edg-l:
The tests need to have the sierra files generated, which is why they fail in CI right now.

I think a good solution would be to add to the test target a call to build the examples, unless you got a better idea. (and also change the ci to call make test instead of cargo)

.github/workflows/ci.yml
```diff
@@ -63,13 +61,12 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: rustfmt
-      - run: cargo fmt --all -- --check
```
edg-l, line 66: this change is wrong, fmt only accepts --all

edg-l (changes requested):

.gitignore
```diff
@@ -9,14 +9,17 @@ Cargo.lock
...
-
 # Added by cargo
-
 /target
 
+# Local LLVM framework.
 llvm/
 
-# Added so we can have our own specific development scripts
-scripts/
```
DomHenderson, line 20: Instead of removing this, could we change it to maybe `dev-scripts/` that is still gitignored? The point of having a place to hold scripts for development that are only useful to the person who wrote them still holds imo even if we now have platform independent ones too

azteca1998: I just tested something and I found another way to keep a local gitignored scripts folder: just having a folder with a `.gitignore` that ignores itself seems to work, and another advantage is that it doesn't have to be named `dev-scripts/`.

What do you think?

edg-l (approved):
lgtm

DomHenderson (changes requested):
felt_div.cairo should be felt_div.cairo.disabled so it doesn't trip up make compile-mlir. Other than that this is great

igaray (approved):

PR #83, opened 2023-04-11
Simplify dataflow libfuncs

DomHenderson:
# Simplifying dataflow libfuncs

## Description

Some libfuncs at the level we care about purely implement dataflow. For example store_temp is just an identity function to us, and since data is never mutated there's no need for dup to actually do anything. As such I introduce a third SierraLibFunc subtype to handle such cases, simplifying output code and making it more readable. This also introduces a change by which there's no need to add libfuncs that are skipped to a list in statement processing

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


edg-l (approved):
lgtm

PR #88, opened 2023-04-12
array get

edg-l:
implements array_get

> due to a possible regression on the order of block arguments not being stable this compiles sometimes only

fixed the regression where block arguments are not in a stable order


examples/index_array.cairo
```diff
@@ -0,0 +1,16 @@
...
+    data.append(4_u8);
+    data.append(5_u8);
+    data.append(4_u8);
+    data.append(4_u8);
+    data.append(4_u8);
+    data.append(1_u8);
+    data.append(1_u8);
+    data.append(1_u8);
+    data.append(2_u8);
+    *data[0]
```
DomHenderson, line 15: To make testing more comprehensive, could this return all the values rather than just the first?

sierra2mlir/src/statements/inline_jumps.rs
```diff
@@ -180,4 +180,119 @@ impl<'ctx> Compiler<'ctx> {
...
+
+            // check if index is out of bounds
+            let cmp_op = self.op_cmp(block, CmpOp::UnsignedLess, index_var.get_value(), length);
+            let cmp = cmp_op.result(0)?.into();
+
+            let block_get_idx = region.append_block(Block::new(&[]));
+
+            // collect args to the panic block
+            let mut args_to_panic_block = vec![];
+            for var_idx in panic_block_info.variables_at_start.keys().sorted() {
```
DomHenderson, line 246: Just a note that if #89 gets in before this then the .sorted() can be removed, and if not then it will have to be updated to handle it

sierra2mlir/src/statements/inline_jumps.rs
```diff
@@ -180,4 +180,119 @@ impl<'ctx> Compiler<'ctx> {
...
+            let length: Value = length_op.result(0)?.into();
+
+            // check if index is out of bounds
+            let cmp_op = self.op_cmp(block, CmpOp::UnsignedLess, index_var.get_value(), length);
+            let cmp = cmp_op.result(0)?.into();
+
+            let block_get_idx = region.append_block(Block::new(&[]));
+
+            // collect args to the panic block
+            let mut args_to_panic_block = vec![];
```
DomHenderson, line 245: No need for mutability here if this is done with a single map expression over `panic_block_info.variables_at_start.keys()` rather than an explicit for loop

DomHenderson (dismissed):

edg-l:
applied the suggestions, but our print differs, sierra wraps functions that may panic in a enum and one of the variants has an array, which for some reason is printed

PR #90, opened 2023-04-13
Add Uninitialized and store_local

DomHenderson:
# Uninitialized<T> and store_local

## Description

Adds uninitialized<T> and store_local. Store local takes two arguments, an uninitialized memory cell and a value, and returns the value (allocated into the memory cell). As far as I can tell, there's no need for us to model this so I've treated it like rename but skipping the first argument. In the case that a user uses an Uninitialized type for something (no idea why they would) I've defined it as the underlying type for cases where we don't omit the underlying type from the program such as RangeCheck.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


edg-l (approved):
lgtm

PR #121, opened 2023-04-17
fix array pop_front and panic panicking when enum is the Ok variant 

edg-l:
- Fixes the memmove on pop_front not using the correct amount of bytes.
- This is the reason a wide mul test was failing, because it introduces a PanicResult enum and it was always constructed with the trap.


sierra2mlir/src/statements/inline_jumps.rs
```diff
@@ -427,8 +427,20 @@ impl<'ctx> Compiler<'ctx> {
             // its safe if new_length is 0
             let new_length_zext_op = self.op_zext(&block_pop_idx, new_length, self.u64_type());
             let new_length_zext = new_length_zext_op.result(0)?.into();
+
+            let element_size_bytes = element_type.get_width() / 8;
```
azteca1998, line 431: I think this might fail if we ever decide to change felts to `i252`.
```suggestion
            let element_size_bytes = (element_type.get_width() + 7) / 8;
```

azteca1998 (approved):
Nice.

PR #122, opened 2023-04-17
implement uint overflow add

edg-l:
work in progress, test is not picking up the output from llvm for some reason, but works on cli


azteca1998 (approved):
Awesome.

sierra2mlir/src/compiler.rs
```diff
@@ -921,15 +899,15 @@ impl<'ctx> Compiler<'ctx> {
     pub fn op_llvm_gep<'a>(
         &self,
         block: &'a Block,
-        index: usize,
+        indexes: &[usize],
```
DomHenderson, line 902: Please correct me if I'm misunderstanding this, but aren't there specifically two indices for gep, one that indexes into the pointer as an array, and one that is to index into the element itself? If so could we have that as two named parameters to this function to avoid confusion?

sierra2mlir/src/utility.rs
```diff
@@ -319,12 +328,12 @@ impl<'ctx> Compiler<'ctx> {
...
-            let enum_data_op =
-                self.op_llvm_extractvalue(&entry_block, 1, enum_value, *storage_type)?;
-            let enum_data = enum_data_op.result(0)?.into();
-            self.op_llvm_store(&entry_block, enum_data, data_ptr)?;
+            //let data_alloca_op = self.op_llvm_alloca(&entry_block, *storage_type, 1)?;
+            //let data_ptr = data_alloca_op.result(0)?.into();
+            //let enum_data_op =
+            //    self.op_llvm_extractvalue(&entry_block, 1, enum_value, *storage_type)?;
+            //let enum_data = enum_data_op.result(0)?.into();
+            //self.op_llvm_store(&entry_block, enum_data, data_ptr)?;
```
DomHenderson, line 336: Is this meant to be left here?

DomHenderson (changes requested):

sierra2mlir/src/compiler.rs
edg-l: No, gep can have infinite (?) indexes and the dialect operation reflects that, by accepting many indices, it's just that when I initially added the gep function we only needed to index into structs, but now we need to index into arrays too, i think this signature is the most faithful to the operation.

https://llvm.org/docs/LangRef.html#getelementptr-instruction
https://mlir.llvm.org/docs/Dialects/LLVM/#llvmgetelementptr-mlirllvmgepop

edg-l: (the dynamic version of the gep which we have (accepts mlir Values) also has this signature)

sierra2mlir/tests/comparison/uint/uint_addition.cairo
```diff
@@ -0,0 +1,29 @@
...
+    (u16, u16, u16),
+    (u32, u32, u32),
+    (u64, u64, u64),
+    (u128, u128, u128),
+) {
+	(
+        (
+            4_u16 + 6_u16,
+            2_u16 + 10_u16,
+            50_u16 + 2_u16,
```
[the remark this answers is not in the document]
edg-l: i can't because wrapping triggers a panicresult which panics, and lli wont output anything

DomHenderson (approved):

PR #124, opened 2023-04-19
Tidy SierraType

DomHenderson:
# Tidy SierraType

## Description

- Moves SierraType into its own file since the impl block was getting quite hefty
- get_array_type -> create_array_type to keep "get" for getting existing data, "create" for creating new data
- More commenting on SierraType methods
- get_width now fails if it the mlir get_type method returns None for a simple type, as every Simple type should have a calculable width
- get_field_types and get_sierra_field_types only return Some for structs, inline with the existing comments. They were only used for structs so far anyway so this doesn't actually change anything other than ensuring consistency between behaviour and comments

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


azteca1998 (approved):
nice

igaray (approved):

PR #131, opened 2023-04-20
Get Or Create helper functions

DomHenderson:
# Get or Create helper functions

## Description

Adds a field to storage to hold the names of helper functions that have been generated. Functions are then generated when the `call_X` method is called for the first time. Also splits up the compiler.rs file into constituent parts since it was getting really big

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


sierra2mlir/src/compiler/external_funcs.rs
```diff
@@ -0,0 +1,197 @@
...
+        self.module.body().append_operation(func);
+
+        Ok(())
+    }
+}
+
+// Function use
+impl<'ctx> Compiler<'ctx> {
+    /// Utility function to create a printf call.
+    /// Null-terminates the string iff it is not already null-terminated
```
edg-l, line 122: typo in the iff

edg-l (approved):

sierra2mlir/src/compiler/external_funcs.rs
igaray: unless it's if and only if

igaray (approved):

PR #138, opened 2023-04-24
merge test and coverage

edg-l:
merges test and coverage since they do nearly the same


DomHenderson (commented):
There's already a codecov.yml in .github/, can they be merged (and ideally kept in the .github folder)?

PR #139, opened 2023-04-24
Refactoring function creation, div fix, array fixes and helpers

DomHenderson:
# TITLE

## Description

Main goal of this PR was to refactor all region handling, and function creation into a single function to be called after the blocks for a function have been created and filled. Along the way, I also:

- Replaced match and generic args with helper
- Added and used more array helpers, which seems to have fixed something as the pop_front_valid test now passes
- Fixed an array append bug where capacity wasn't multiplied by the size of the element before being passed to realloc
- Reimplemented felt_div with the extended euclidean algorithm which should be faster than exponentiation. Div tests now pass
- Replaced Block::new with self.new_block as it's more concise and easier to read
- Replaced large `if let`s with smaller `let (x,y) = if let`, both to reduce scope nesting and to prepare for SierraType refactors that'll replace such things with a function call
- Fixed a segfault on benches and when executing a function that returns a value.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


edg-l:
Merged main here and fixed conflicts, also some test issues

azteca1998 (approved):

Issue #66, opened 2023-03-30
Libfunc: gas


igaray:
- Compiler internals:
  - [x] Gas global variable
  - [x] Gas global variable manipulation (increase, decrease, get) 
- Libfuncs:
  - [x] withdraw_gas
  - [x] withdraw_gas_all
  - [x] get_available_gas
  - [x] get_builtin_costs

Some documentation references: 

- [ StarkWare Sessions 23 | Sierra - Enforcing Safety Using Typesystems | Shahar Papini ](https://www.youtube.com/watch?v=-EHwaQuPuAA)
- [ StarkWare Sessions 23 | Not Stopping at the Halting Problem | Ori Ziv ](https://www.youtube.com/watch?v=wYxUedcdVZ4&list=PLcIyXLwiPilV7OFs_bkGe1-qqE3Y4QiEl&index=17)
- [StarkNet Documentation: Fee Mechanism](https://docs.starknet.io/documentation/architecture_and_concepts/Fees/fee-mechanism/)

> Note: Cairo prevents us from running program with infinite loops by including a gas meter. The gas meter is a mechanism that limits the amount of computation that can be done in a program. By setting a value to the --available-gas flag, we can set the maximum amount of gas available to the program. Gas is a unit of measurementsthat expresses the computation cost of an instruction. When the gas meter runs out, the program will stop. In this case, the program panicked because it ran out of gas, as the stop condition was never reached. It is particularly important in the context of smart contracts deployed on Starknet, as it prevents from running infinite loops on the network. If you're writing a program that needs to run a loop, you will need to execute it with the --available-gas flag set to a value that is large enough to run the program.

- [src](https://cairo-book.github.io/ch02-05-control-flow.html?highlight=gas#repeating-code-with-loop)
- [withdraw gas example](https://cairo-book.github.io/ch05-01-enums.html?highlight=gas#the-option-enum-and-its-advantages)

> As recursive functions can be a source of infinite loops if the stop condition is never met, the Cairo-to-Sierra compiler will inject a withdraw_gas method at the beginning of recursive functions. As this feature has not yet been implemented, developers are still expected to call withdraw_gas in recursive functions and handle the result themselves, although it should be included in a future release of the compiler.
[...]

PR #140, opened 2023-04-25
initial gas work

edg-l:
- Added gas counter global.
- Added helper functions to modify the gas global.
- Added types: gasbuiltin, builtincosts.
- Added gas user function gas calculation, used to reduce current gas on function calls.
- Implemented builtins: get_available_gas, withdraw_gas, withdraw_gas_all, branch_align, get_builtin_costs
- Added test
- Disabled example/uint due to mul guarantee not being implemented. (introduced in cairo rc0)

Related to #66 

There are still places where gas is deducted but i have to find them yet.


sierra2mlir/tests/comparison.rs
```diff
@@ -23,52 +23,53 @@ use test_case::test_case;
 // Tests behaviour of the generated MLIR against the behaviour of starkware's own sierra runner
 // Such tests must be an argumentless main function consisting of calls to the function in question
 
-#[test_case("array/append")]
-#[test_case("array/index_invalid")]
-#[test_case("array/pop_front_invalid")]
+#[test_case("array/append", None)]
```
DomHenderson, line 26: Gas tracking feels like something that needs thorough testing, would it be possible (and do you think it's a good idea) to run all tests that do not depend on it without gas, and then also run them with some large amount of available gas and check the correct amount is used?

sierra2mlir/src/statements/mod.rs
```diff
@@ -381,6 +381,16 @@ impl<'ctx> Compiler<'ctx> {
                 .map(|t| (t.ty.get_type(), Location::unknown(&self.context)))
                 .collect_vec(),
         );
+
+        if let Some(gas) = &self.gas {
+            // reduce the gas cost of this user function call
```
DomHenderson, line 386: If I'm inferring intent correctly I think this should read "spend the required gas for the user function call"/"reduce the counter by..." etc, as it's not reducing the cost of the function call itself

sierra2mlir/src/compiler/mod.rs
```diff
@@ -44,7 +67,56 @@ impl<'ctx> Compiler<'ctx> {
...
+        available_gas: Option<usize>,
     ) -> color_eyre::Result<Self> {
+        let sierra_program_registry = ProgramRegistry::<CoreType, CoreLibfunc>::new(program)?;
+
+        // Create the compiler gas structure if gas was provided, similar to cairo casm runner.
+        let gas: Option<CompilerGas> = if let Some(available_gas) = available_gas {
+            let function_set_costs: OrderedHashMap<FunctionId, OrderedHashMap<CostTokenType, i32>> =
+                Default::default();
+
+            let pre_function_set_costs = function_set_costs
```
DomHenderson, line 79: This whole block could probably do with some step by step comments as to what each of these maps is for and why it's collected the way that it is. It's a little difficult to process as is

edg-l: i mostly copied this from the cairo-runner crate

sierra2mlir/tests/comparison.rs
edg-l: i can enable it for tests that don't use loops or recursion, this is due to them not working with gas enabled because they need to use withdraw_gas with a match to check for gas availability, you can try it with cairo-run --available-gas 5000 and one of the fib examples.

edg-l:
implemented withdraw_gas and branch_align

igaray (approved):

PR #159, opened 2023-05-20
[readme] add ninja requirement and update setup instructions

whymidnight:
the readme didn't include the ninja requirement for cmake and untar yielded without `.tar`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ make book
 - mdbook
 - LLVM 16+ with MLIR
 - Rust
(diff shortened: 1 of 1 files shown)


edg-l:
In the README we list the hard dependencies of sierra2mlir, ninja is a optional build tool you can tell cmake to use, so we would rather not add it there.

Regarding the README instructions, can you update the llvm url to the new version 16.0.4? Thanks

edg-l:
closing as it's stale and it's no longer meaningful

PR #185, opened 2023-06-22
fix refactor build + misc

edg-l:
# TITLE

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/libfuncs/uint64.rs
```diff
@@ -1 +1,112 @@
...
+    let value = info.c;
+    let value_ty = registry
+        .get_type(&info.signature.branch_signatures[0].vars[0].ty)
+        .unwrap()
+        .build(context, helper, registry, metadata)
+        .unwrap();
+
+    let op0 = entry.append_operation(arith::constant(
+        context,
+        IntegerAttribute::new(value.try_into().unwrap(), value_ty).into(),
```
azteca1998, line 71: This will fail when the constant value is greater than `isize::MAX`. I think we should use the format string here.
```suggestion
        Attribute::parse(context, &format!("{value} : {value_ty}").unwrap(),
```

src/types.rs
```diff
@@ -142,13 +142,13 @@ impl TypeBuilder for CoreTypeConcrete {
...
             CoreTypeConcrete::Felt252(_) => get_integer_layout(252),
             CoreTypeConcrete::GasBuiltin(_) => get_integer_layout(64),
-            CoreTypeConcrete::BuiltinCosts(_) => todo!(),
-            CoreTypeConcrete::Uint8(_) => todo!(),
-            CoreTypeConcrete::Uint16(_) => todo!(),
-            CoreTypeConcrete::Uint32(_) => todo!(),
-            CoreTypeConcrete::Uint64(_) => todo!(),
-            CoreTypeConcrete::Uint128(_) => todo!(),
-            CoreTypeConcrete::Uint128MulGuarantee(_) => todo!(),
+            CoreTypeConcrete::BuiltinCosts(_) => Layout::new::<()>(),
```
azteca1998, line 145: Is this builtin type really `()`? Shouldn't this be a pointer to a cost table or something?

src/types.rs
```diff
@@ -142,13 +142,13 @@ impl TypeBuilder for CoreTypeConcrete {
...
+            CoreTypeConcrete::Uint8(_) => get_integer_layout(8),
+            CoreTypeConcrete::Uint16(_) => get_integer_layout(16),
+            CoreTypeConcrete::Uint32(_) => get_integer_layout(32),
+            CoreTypeConcrete::Uint64(_) => get_integer_layout(64),
+            CoreTypeConcrete::Uint128(_) => get_integer_layout(128),
+            CoreTypeConcrete::Uint128MulGuarantee(_) => Layout::new::<()>(),
```
azteca1998, line 151: Same here, I think we should match the vm's behavior even though we won't really end up using this type for anything. Or at least have the option to, with a flag or something. What do you think?

src/types.rs
```diff
@@ -192,7 +192,7 @@ impl TypeBuilder for CoreTypeConcrete {
             CoreTypeConcrete::Poseidon(_) => todo!(),
             CoreTypeConcrete::Span(_) => todo!(),
             CoreTypeConcrete::StarkNet(_) => todo!(),
-            CoreTypeConcrete::SegmentArena(_) => todo!(),
+            CoreTypeConcrete::SegmentArena(_) => Layout::new::<()>(),
```
azteca1998, line 195: I'm not sure what `SegmentArena` stores. Maybe the same could be applied here?

azteca1998 (approved):
Nice.

PR #187, opened 2023-06-27
implement snapshot type and more libfuncs

edg-l:


src/libfuncs/snapshot_take.rs
```diff
@@ -1,3 +1,45 @@
...
+    _info: &SignatureOnlyConcreteLibfunc,
+) -> Result<(), std::convert::Infallible>
+where
+    TType: GenericType,
+    TLibfunc: GenericLibfunc,
+    <TType as GenericType>::Concrete: TypeBuilder,
+    <TLibfunc as GenericLibfunc>::Concrete: LibfuncBuilder,
+{
+    // TODO: Should this act like dup like it does now? or are there special requirements. So far it seems to work.
+    // TODO: Handle non-trivially-copyable types (ex. arrays) and maybe update docs.
```
azteca1998, line 33: Judging by the name and the fact that there are references (and are different from snapshots) maybe a `Snapshot<T>` is roughly equivalent to a `Cow<T>` where it provides a "copy" of a value that is not really copied until used/taken, but doesn't get updated if the original changes?

azteca1998 (approved):

PR #192, opened 2023-06-29
fix ci + array len libfunc

edg-l:
fixes ci and adds the array len libfunc


src/libfuncs/uint8.rs
```diff
@@ -275,8 +275,8 @@ mod test {
     use crate::utils::test::run_cairo;
     use serde_json::json;
 
-    fn error_value() -> serde_json::Value {
-        json!([(), [1, [[], [608642104203229548495787928534675319u128]]]])
+    fn error_value(n: u128) -> serde_json::Value {
+        json!([(), [1, [[], [n]]]])
     }
```
azteca1998, line 280: The idea of this helper function is to not have to type the huge number every time. Adding the number as a parameter defeats the purpose.

Maybe add two separate functions?

```rs
fn u8_add_overflow_error() { ... }
fn u8_sub_overflow_error() { ... }
```

src/utils.rs
```diff
@@ -171,6 +166,11 @@ pub mod test {
             .run(&mut module)
             .expect("Could not apply passes to the compiled test program.");
 
+        assert!(
+            module.as_operation().verify(),
+            "Test program generated invalid MLIR."
+        );
+
```
azteca1998, line 173: I think this can be moved back after the fix I made.

azteca1998 (approved):
Nice.

PR #201, opened 2023-07-03
Array pop front and pop front consume

edg-l:
adds more array libfuncs

## Checklist
- [x] Unit tests added


src/libfuncs/array.rs
```diff
@@ -59,8 +59,13 @@ where
...
             build_append(context, registry, entry, location, helper, metadata, info)
         }
-        ArrayConcreteLibfunc::PopFront(_) => todo!(),
-        ArrayConcreteLibfunc::PopFrontConsume(_) => todo!(),
+        ArrayConcreteLibfunc::PopFront(info) => {
+            build_pop_front(context, registry, entry, location, helper, metadata, info)
+        }
+        ArrayConcreteLibfunc::PopFrontConsume(info) => {
+            // same signature at sierra level
+            build_pop_front(context, registry, entry, location, helper, metadata, info)
```
azteca1998, line 67: Would it be better to create the function and forward it from there? In this way, if we ever need to search the libfunc we'll be able to find it by its name.

azteca1998 (dismissed):
Awesome.

azteca1998 (dismissed):

igaray (dismissed):

azteca1998 (approved):

igaray (approved):

PR #207, opened 2023-07-04
Bool libfuncs and bench fixes

edg-l:


src/libfuncs/felt252.rs
```diff
@@ -345,6 +355,13 @@ mod test {
         u32_digits.try_into().unwrap()
     }
 
+    #[test]
+    fn felt252_factorial() {
+        let r = |n| run_program(&FELT252_FACTORIAL, "factorial", json!([null, 0, n]));
+
+        assert_eq!(r(f("5")), json!([null, 0, [0, [f("120")]]]));
+    }
+
```
azteca1998, line 364: Same as with the other PR, I think we should move this into the integration tests.

src/libfuncs/bool.rs
```diff
@@ -1,3 +1,316 @@
...
+
+    #[test]
+    fn run_not() {
+        let program = load_cairo!(
+            use array::ArrayTrait;
+
+            fn run_test() -> bool {
+                true == true
+            }
+        );
```
azteca1998, line 225: Calling `load_cairo!` is extremely expensive (that was the whole point in changing the way testing works). Maybe you could use parameters instead of constants and keep one `load_cairo!` invocation per test? Or even better, one invocation per multiple tests.

azteca1998 (dismissed):
Nice.

src/libfuncs/felt252.rs
edg-l: i'll move this in a following pr

azteca1998 (approved):

igaray (approved):

PR #220, opened 2023-07-10
Enum comparision tests

edg-l:
adds tests for enums

### DIFF
--- a/tests/common.rs
+++ b/tests/common.rs
@@ -1,5 +1,5 @@
 #![allow(unused_macros)]
-#![allow(unused)]
+#![allow(dead_code)]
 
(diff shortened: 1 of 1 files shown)


tests/common.rs
```diff
@@ -1,4 +1,5 @@
 #![allow(unused_macros)]
+#![allow(unused)]
```
azteca1998, line 2: This is dangerous. I think it would be better to no have it at all, but if it's really needed, then:
```suggestion
#![allow(dead_code)]
```

azteca1998 (approved):

azteca1998 (approved):

igaray (approved):

PR #230, opened 2023-07-13
dict libfuncs progress

edg-l:
Libfuncs done:

- alloc_local
- felt252_dict_entry_finalize
- felt252_dict_entry_get
- felt252_dict_new
- felt252_dict_squash
- finalize_locals
- store_local

Serialize and deserialize works.


src/values.rs
```diff
@@ -117,9 +117,9 @@ impl ValueBuilder<CoreType, CoreLibfunc> for CoreTypeConcrete {
             CoreTypeConcrete::Uninitialized(_) => todo!(),
             CoreTypeConcrete::Enum(_) => true,
             CoreTypeConcrete::Struct(_) => true,
-            CoreTypeConcrete::Felt252Dict(_) => todo!(),
+            CoreTypeConcrete::Felt252Dict(_) => true,
             CoreTypeConcrete::Felt252DictEntry(_) => todo!(),
-            CoreTypeConcrete::SquashedFelt252Dict(_) => todo!(),
+            CoreTypeConcrete::SquashedFelt252Dict(_) => true,
```
azteca1998, line 122: They are pointers (aka. `u64`), therefore not complex.
```rs
            CoreTypeConcrete::Felt252Dict(_) => false,
            CoreTypeConcrete::Felt252DictEntry(_) => todo!(),
            CoreTypeConcrete::SquashedFelt252Dict(_) => false,
```

src/utils.rs
```diff
@@ -175,6 +176,46 @@ macro_rules! codegen_ret_extr {
 }
 pub(crate) use codegen_ret_extr;
 
+pub fn register_runtime_symbols(engine: &ExecutionEngine) {
+    #[cfg(feature = "with-runtime")]
+    unsafe {
```
azteca1998, line 181: I think this function may be better as suggested below because if the `with-runtime` feature is disabled it doesn't make sense to have a `register_runtime_symbols` function.

```suggestion
#[cfg(feature = "with-runtime")]
pub fn register_runtime_symbols(engine: &ExecutionEngine) {
    unsafe {
```

Changing this will require adding `#[cfg(features = "with-runtime")]` to the function's invocations.

azteca1998 (dismissed):

azteca1998 (approved):

igaray (approved):

PR #237, opened 2023-07-19
implement nullable libfuncs

edg-l:
- [x] `match_nullable`
- [x] `null`
- [x] `nullable_from_box`

depends on #236


src/values/nullable.rs
```diff
@@ -1 +1,164 @@
...
+impl<'a, 'de, TType, TLibfunc> de::Visitor<'de> for Visitor<'a, TType, TLibfunc>
+where
+    TType: GenericType,
+    TLibfunc: GenericLibfunc,
+    <TType as GenericType>::Concrete: TypeBuilder<TType, TLibfunc> + ValueBuilder<TType, TLibfunc>,
+{
+    type Value = NonNull<()>;
+
+    fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "An array of values with the same type.")
```
azteca1998, line 105:
```suggestion
        write!(f, "An optional value.")
```

igaray (approved):

PR #245, opened 2023-07-24
tests with automatic comparision against cairo run, first part

edg-l:
Adds more comparision tests for each libfunc while implementing an automatic comparation of the results between cairo-native (JIT) and cairo-runner (from starkware)

There are still lot more tests that need to be added, but they can come in another PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ check:
 test:
 	cargo test --profile optimized-dev --all-targets --all-features
 
-proptests:
+proptest:
 	cargo test --profile optimized-dev --all-targets --all-features proptest
 
 coverage:
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,9 @@ check:
 test:
 	cargo test --profile optimized-dev --all-targets --all-features
 
+proptests:
+	cargo test --profile optimized-dev --all-targets --all-features proptest
+
 coverage:
 	cargo llvm-cov --profile optimized-dev --all-features --workspace --lcov --output-path lcov.info
 
--- a/tests/common.rs
+++ b/tests/common.rs
@@ -258,11 +258,49 @@ pub fn compare_outputs(
                 }
             }
             CoreTypeConcrete::BuiltinCosts(_) => todo!(),
-            CoreTypeConcrete::Uint8(_) => todo!(),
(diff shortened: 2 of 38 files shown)


Makefile
```diff
@@ -21,6 +21,9 @@ check:
 test:
 	cargo test --profile optimized-dev --all-targets --all-features
 
+proptests:
```
azteca1998, line 24: I think maybe keeping it in singular is more in line with the rest of the rules (ex. `test` vs `tests`).
```suggestion
proptest:
```

tests/common.rs
```diff
@@ -184,17 +209,333 @@ pub fn run_native_program(
...
+    for ty in ret_types {
+        let ty = reg.get_type(ty).unwrap();
+        check_next_type(ty, ignore_gas, &mut native_rets, &mut vm_rets, vm_gas, &reg)?;
+    }
+
+    Ok(())
+}
+
+pub const FIELD_HIGH: u128 = (1 << 123) + (17 << 64); // this is equal to 10633823966279327296825105735305134080
+pub const FIELD_LOW: u128 = 1;
```
azteca1998, line 507: I think this is `felt252`'s prime modulo, which we have available at `cairo_native::types::felt252::PRIME` if I'm not mistaken.

tests/common.rs
```diff
@@ -184,17 +209,333 @@ pub fn run_native_program(
...
         args,
         gas,
         StarknetState::default(),
     )
 }
+
+/// Given the result of the cairo-vm and cairo-native of the same program, it compares
+/// the results automatically, triggering a proptest assert if there is a mismatch.
+///
+/// If ignore_gas is not false, it will check whether the resulting gas matches.
```
azteca1998, line 231: Does this mean that if `ignore_gas == true` then it won't ignore it?

tests/common.rs
edg-l: i guess, but the high and low fields are used in the proptest value generation

i mainly copied it from the cairo_felt crate, which has a feature for this, but because cairo-vm uses an older version which doesn't have the feature, i had to copy this over.

azteca1998 (approved):
Awesome.

igaray (approved):

PR #255, opened 2023-07-31
Overall improvements to the look of the project

edg-l:
This PR adds the following:

- Improved revamped README:
- - Improved first impression.
- - The libfunc list is now under a collapsible dropdown, the users who want to read it can simply click on it, it's still among the first items, but if you want to scroll past it, it won't be as burdensome now.
- - Listed and added more make targets. New make targets: doc, doc-open, install, build-native
- - Added mlir resources
- - Improved explanation of the list of libfuncs and what they are.
- **The Makefile now automatically fetches the corelibs if they are missing, thus making the process more seamless.**
- Added code owners file.
- Added more metadata to the `Cargo.toml` file
- Removed `Cargo.lock` from git ignores, since we have binaries and thus it shouldn't be git ignored according to rust guidelines.
- Added some more docs in the code.
- Removed mentions to mdbook, which will be unused for now.
- Added rustdoc to CI and gh-pages.
- Made the release version use LTO and lower codegen units for maximum perfomance, as stated in the [rust perfomance book](https://nnethercote.github.io/perf-book/build-configuration.html)
- Slightly edited the dockerfile to use the locked cargo.lock deps, to avoid issues in the future in case some minor update breaks.

image

Check the rendered markdown: https://github.com/lambdaclass/cairo_native/tree/improve_docs_edgar


unbalancedparentheses (approved):

unbalancedparentheses:
Let's merge this since it's way better than before. We can improve it afterwards and review everything. It´s a readme.

pablodeymo (approved):

igaray (approved):

PR #305, opened 2023-08-28
Make StarknetSyscallHandler trait receive mut self for integration with SiR

entropidelic:
# TITLE
In order to modify state, some StarknetSyscallHandler trait methods have to receive a mutable reference to self.
Also, this PR removes the PhantomData used in the SyscallHandlerMeta for the lifetime

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/examples/erc20.rs
+++ b/examples/erc20.rs
@@ -88,7 +88,7 @@ impl StarkNetSyscallHandler for SyscallHandler {
     }
 
     fn storage_read(
-        &self,
+        &mut self,
         address_domain: u32,
         address: cairo_felt::Felt252,
     ) -> SyscallResult<cairo_felt::Felt252> {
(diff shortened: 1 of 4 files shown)


jrchatruc (approved):

azteca1998 (commented):
I think changing all the methods' `&self` to `&mut self` is in order. It was meant to be this way sooner or later anyways.

The lifetime change is more important, especially when used as a library, as there will be no guarantee that the stored pointer in the metadata won't end up as a dangling pointer, or worse, pointing to different stuff within the stack and corrupting the entire program. If the lifetime cannot be restored, I'd suggest maybe marking marking it as unsafe and leaving a warning in the docs.

igaray (approved):

Issue #328, opened 2023-10-26
Bug when with Err variant on results from a Starknet syscall


edg-l:
Check out https://github.com/lambdaclass/cairo_native/pull/327

While doing https://github.com/lambdaclass/starknet_in_rust/pull/1102 i found a bug, the felt error was not there and the deserialization crashed (its now fixed with a deserialization error, but it should find the felt because its passed)

When a syscall returns a Err (that always contains a vec of felts) it wont see the vec of felts but they are there.

https://github.com/lambdaclass/cairo_native/blob/120f972724ebe5dafce4f7a7b766551187067148/src/libfuncs/stark_net.rs#L1281-L1381

I think the problem here is how we handle the return syscall payload in the starknet libfuncs, when the return type is () we use a hack to get the enum type have 2 variants (but we cant make a concretetypeid from nowhere so we copy the variant from the other err branch)
We also build the ok and err payload always regardless of whether the enum tag is 0 or 1 for each variant, i think the problem lies there (edited)

PR #335, opened 2023-10-30
Fix `llvm.getelementptr` offset bug.

azteca1998:
# Fix `llvm.getelementptr` offset bug

## Description

We were computing a byte offset on a pointer taking into account the actual type, causing the offset to be off by a factor of the actual type's length.

In other words, we were calculating the offset as if it were a C pointer, whereas we needed a byte offset. Example:
```c
my_type_t *ptr;

// Wrong:
my_type_t *offset_ptr = ptr + offset;

//  Right:
my_type_t *offset_ptr = (my_type_t *) (((void *) ptr) + offset);
```

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/stark_net.rs
+++ b/src/libfuncs/stark_net.rs
@@ -493,7 +493,7 @@ where
                         ),
                         (
                             Identifier::new(context, "elem_type"),
-                            TypeAttribute::new(variant_tys[1].0).into(),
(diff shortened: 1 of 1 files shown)


edg-l:
fixes #328

edg-l (approved):

PR #342, opened 2023-11-13
Make the context `Send + Sync`.

azteca1998:
# Make the context `Send + Sync`

## Description

Depends on #343.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/bin/cairo-native-run.rs
```diff
@@ -130,6 +130,11 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     #[cfg(feature = "with-runtime")]
     register_runtime_symbols(&engine);
+    #[cfg(feature = "with-debug-utils")]
+    metadata
+        .get::<cairo_native::metadata::debug_utils::DebugUtils>()
+        .unwrap()
```
juanbono, line 136: is it ok to unwrap here?

juanbono (approved):
I left a small comment but otherwise looks good to me.

src/bin/cairo-native-run.rs
azteca1998: Yes, there are various reasons:
  - We don't have an error for missing metadata yet. You'll notice that every time we need some metadata there's an unwrap for now.
  - The real reason: When `#[cfg(feature = "with-debug-utils")]` is set, the debug utils [are automatically inserted](https://github.com/lambdaclass/cairo_native/blob/main/src/metadata.rs#L106-L117) when the metadata storage is created. In other words, the `#[cfg(feature = "with-debug-utils")]` ensures that the unwrap will never fail (of course, assuming we don't remove it manually, which we shouldn't).

edg-l (approved):

PR #354, opened 2023-11-27
aot library load

edg-l:
## What this PR does

Using the libloading crate, we load the compiled library and call the entrypoint using the `_mlir_cifafe_` wrapper that gives us a `extern "C"` function callable from Rust side.

## The problem

The problem is figuring out the correct signature and how to pass the correct arguments.

For now I reduced the scope to just calling contract function wrappers, which are produced by cairo->sierra when compiling a starknet contract, they always have the same signature:

```mlir
llvm.func @"_mlir_ciface_hello_starknet::hello_starknet::Echo::__wrapper__echo"
        (%arg0: !llvm.ptr, // this argument is the returned value behind a pointer (AFAIK), so its what will be on -> X at the rust side.
         %arg1: !llvm.array<0 x i8>,  // This is the range check builtin, zero sized.
         %arg2: i128, // The gas
         %arg3: !llvm.ptr, // The syscall handler
         %arg4: !llvm.struct<(struct<(ptr<i252>, i32, i32)>)>) // The Span<Felt252>, AKA the call data, to my knowledge a span is a struct whose first field is a array.
        attributes {llvm.emit_c_interface, sym_visibility = "public"} {}
```

I still havent figured out how to give the arguments correctly, right now i re-use the JIT value to construct the values:

```rust
       let calldata = JITValue::Struct {
            fields: vec![JITValue::Array(vec![
                JITValue::Felt252(1.into()),
            ])],
            debug_name: None,
        }
        .to_jit(&arena, reg, ty)
        .unwrap();

        dbg!(&calldata);

        let syscall_handler_meta = SyscallHandlerMeta::new(syscall_handler);

        let syscall_addr = syscall_handler_meta.as_ptr().as_ptr() as *const () as usize;

        let syscall_alloc = arena.alloc(syscall_addr as *mut ());
        let return_value = libc::malloc(8094);
        return_value.cast::<u32>().write(0xbeef);

        let gas_ptr: *mut u128 = arena.alloc_layout(Layout::new::<u128>()).as_ptr().cast();
        gas_ptr.write(u64::MAX.into());

        let func: libloading::Symbol<
            unsafe extern "C" fn(
                return_value: *mut (),
                range_check: *const (),
                gas_builtin: *mut u128,
                syscall_handler: *mut (),
                calldata: *mut (),
            ) -> *mut RetValue,
        > = lib.get(format!("_mlir_ciface_{}", symbol).as_bytes())?;

        let gas: u128 = u64::MAX.into();
        let range_check = arena.alloc_layout(Layout::new::<()>()).as_ptr().cast();

        // segfaults due to a memmove on the calldata pointer (?)
        let result = func(
            return_value.cast(),
            range_check,
            gas_ptr,
            syscall_alloc.cast(),
            calldata.as_ptr().cast(),
        );
```

~~But doesn't seem to work out well, because the later code calls a pop_front on the calldata we pass, and somehow when loading the first element, it SEGFAULTS, the part creating the segfault is here:~~

https://github.com/lambdaclass/cairo_native/blob/7a5a006bb904dfce650881936cf143b8e420e209/src/libfuncs/array.rs#L694-L704

Edit: the segfault is no longer there after providing with 1 more argument, it seems the calldata pointer is picked up correctly, but there is a memmove involved in the pop_front that causes a segfault, probably due to misalignment (?) or it could also be because the pointer is not from libc, although i think the array is allocated using rust libc crate, but there may be issued if both libc versions arent exactly the same?.

## macOS

On macOS i didn't get the segfault and i got to print some values returned, but they didn't make much sense, i think its also because the arguments are passed wrongly but somehow the execution went through. I also did not see any prints from the syscall handler which means something went wrong there too.

# Reproducing

Testing is done using the aot.rs example.

You will need to change some lines because right now to use the cairo_runtime library from this same repo i use hardcoded paths, you can change them here:

https://github.com/lambdaclass/cairo_native/blob/7a5a006bb904dfce650881936cf143b8e420e209/src/ffi.rs#L154-L187

To run:

```bash
cargo b --all-features --all-targets --examples

# Then to debug i use rust-gdb on linux and rust-lldb on macos:
rust-gdb  ./target/debug/examples/aot
```


edg-l:
Closing as #363 supersedes this

PR #363, opened 2023-12-05
Cairo AOT preview.

azteca1998:
# Cairo AOT preview

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/aot.s
```diff
@@ -0,0 +1,69 @@
...
+    cmp     x2,     0                       // If there are no arguments,
+    beq     3f                              // skip argument processing completely.
+
+    //
+    // Preparation.
+    //
+    add     x1,     x1,     x2,     lsl 3   // Move the pointer to the end (past last element).
+
+    str     x19,    [sp, #-8]!              // Push the x19 register (we'll use it to store sp).
+    mov     x19,    sp                      // Store the original stack pointer for later.
```
Oppen, line 22: I think we need to align the stack to 16 bytes. Maybe that's why we get `SIGBUS` for unaligned access.

src/aot.s
```diff
@@ -0,0 +1,69 @@
...
+    ldr     x1,     [x10, #-8]!             // Decrement pointer, then load the value.
+    ldr     x0,     [x10, #-8]!             // Decrement pointer, then load the value.
+
+  3:
+    // Call the function.
+    blr     x0
+
+    // Restore the original stack pointer and x19 register.
+    mov     sp,     x19
+    ldr     x19,    [sp],   8
```
Oppen, line 67: Is this also correct for the zero arguments case? We didn't touch this in that case.

src/aot.s
azteca1998: The 16-byte requirement is for x86. I'm not aware of ARM having that requirement.

src/aot.s
azteca1998: No, it's not correct. Neither is `blr x9` when calling without arguments.

src/aot.s
```diff
@@ -0,0 +1,69 @@
...
+    add     x1,     x1,     x2,     lsl 3   // Move the pointer to the end (past last element).
+
+    str     x19,    [sp, #-8]!              // Push the x19 register (we'll use it to store sp).
+    mov     x19,    sp                      // Store the original stack pointer for later.
+
+    //
+    // Copy stack arguments (for n_args > 8).
+    //
+    cmp     x2,     8                       // Check if there are more than 8 arguments.
+    ble     2f                              // If there are less than 8, skip to register arguments.
```
Oppen, line 28: But this would be incorrect when one argument <8 is an enum, right? Or do we report them padded?

azteca1998: Yes, assuming we passed the arguments as-is to the trampoline. The idea was to prepare them in Rust and send them to the trampoline to move them where they need to be. That's why it's an array of `u64` instead of `JitValue`.

src/aot.s
Oppen: Maybe I'm misreading here, but I think Apple also puts that requirement: https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Pass-arguments-to-functions-correctly

src/aot.s
Oppen: But we jump here from line 14 in that case, so this is a bug, right?
Also, now that you mention it, I think there's a typo in line 63. It branches to `x0`, which no longer contains the function pointer. As you said, you should branch to `x9`, not `x0`, right?

src/aot.s
azteca1998: Oh you're right. I'm pretty sure I did search that link for alignment stuff but it seems I missed it or searched a different page. Thanks.

src/aot.s
azteca1998: Yes, I've fixed it. It now works for function calls without stack. I still need to fix the stack alignment.

src/executor/aot.rs
```diff
@@ -0,0 +1,367 @@
...
+            );
+        }
+
+        // Parse return values.
+        let type_info = self
+            .registry
+            .get_type(function_signature.ret_types.last().unwrap())
+            .unwrap();
+        let return_value = match type_info {
+            CoreTypeConcrete::Array(_) => match return_ptr {
```
edg-l, line 131: I think you also need to handle the snapshot return type, which is basically the inner type. On the from_jit we do this

`CoreTypeConcrete::Snapshot(info) => Self::from_jit(ptr, &info.ty, registry),`

but i think you should also add this case here. Afaik snapshots can only be made for arrays (maybe structs but i havent seen it)

PR #376, opened 2023-12-12
Implement `Sint8` type + libfuncs

fmoletta:
Most Sint8 libfuncs are the exact same as their Uint8 counterparts with small modifications to wide_mul and operation, and the addition of the diff libfunc


edg-l (approved):
I want to mention that you don't need to store the operation as a variable to get its result, we used to do this in the past due to lifetime issues but with the current melior version it's not needed anymore (and you will see that most new code uses the second form i show).
```rust
let op = entry.append_operation(arith::extui(value, felt252_ty, location));
let result = op.result(0)?.into();
```

```rust
let result = entry.append_operation(arith::extui(value, felt252_ty, location)).result(0)?.into();
```

PR #386, opened 2023-12-14
Fix bug in `ec_point_from_x`

fmoletta:
Libfunc `ec_point_from_x_nz` now uses -y instead of y when y > PRIME / 2
Add comparison between native and vm values for `NonZero` types
Tests `ec_point_from_x_proptest`, `ec_point_from_x_big` &  `ec_point_from_x_small` now passing


edg-l:
Thanks for fixing this bug!

edg-l (approved):

PR #392, opened 2023-12-18
document todos and implement bytes31

edg-l:
Fixes #377
Fixes #391

Description of the pull request changes and motivation.

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added
- [ ] Integration tests added.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.

PR #394, opened 2023-12-18
Add tests using `alexandria` cairo library

fmoletta:
Based on #392 (As bytes32 libfuncs are needed for the test)


.github/workflows/ci.yml
```diff
@@ -134,6 +134,10 @@ jobs:
         uses: dtolnay/rust-toolchain@1.72.1
         with:
           components: clippy
+      - name: Install scarb 
+        uses: software-mansion/setup-scarb@v1
+        with:
+          scarb-version: "2.4.0"
```
Oppen, line 140: I'd suggest we just use the `make` target. We'll eventually want to do proper scrutiny of external actions so being conservative about new ones sounds like a good idea to me.
@klaus993 @pefontana what do you think?

Oppen (approved):

.github/workflows/ci.yml
fmoletta: The `deps` target installs scarb using the installation script, but I couldn't get it to work for the macos CI, thats why I had to add the actions. We can create an issue to remove the action once the installation is fixed

edg-l (approved):

PR #397, opened 2023-12-19
Use function id number when generating function names to avoid repetitions in generic functions

fmoletta:
When compiling a cairo program with generic functions, the sierra may "hide" the generic function's parameter's types. This means that multiple versions of a generic function may show up with the same name in the generated sierra. For example this is the sierra generated by the [cairo file added in this PR](https://github.com/lambdaclass/cairo_native/blob/generic-fn-loop-bug/tests/cases/generic_fn_loop.cairo):
```
function_call<user@generic_fn_loop::generic_fn_loop::bar[expr0]>() -> ([0]); // 0
drop<Unit>([0]) -> (); // 1
function_call<user@generic_fn_loop::generic_fn_loop::bar[expr0]>() -> ([1]); // 2
drop<Unit>([1]) -> (); // 3
u16_const<3>() -> ([2]); // 4
u64_const<5>() -> ([3]); // 5
struct_construct<Tuple<u16, u64>>([2], [3]) -> ([4]); // 6
store_temp<Tuple<u16, u64>>([4]) -> ([5]); // 7
return([5]); // 8
struct_construct<Unit>() -> ([0]); // 9
store_temp<Unit>([0]) -> ([1]); // 10
return([1]); // 11
struct_construct<Unit>() -> ([0]); // 12
store_temp<Unit>([0]) -> ([1]); // 13
return([1]); // 14
```
 When native compiles this, there are two functions with the `generic_fn_loop::generic_fn_loop::bar[expr0]` debug name, leading to an invalid mlir being generated, as both functions take the same name.
 This PR fixes this problem by adding the function's id number when generating its name, therefore the two functions named `generic_fn_loop::generic_fn_loop::bar[expr0]` become `generic_fn_loop::generic_fn_loop::bar[expr0](f1)` & `generic_fn_loop::generic_fn_loop::bar[expr0](f2)`


azteca1998 (approved):
Nice find.

Oppen (dismissed):

igaray (dismissed):

igaray (approved):

edg-l (approved):

PR #417, opened 2024-01-11
Sandboxing

edg-l:
Design:

- A new IsolatedExecutor which spawns the cairo-executor binary and manages it, sending and receiving messages via ipc
- A new binary called cairo-executor which is spawned by IsolatedExecutor and receives and sends messages via ipc, executing programs and returning the results. It also forwards the syscalls via ipc, so ultimately the syscallhandler is implemented by the host side. The cairo-executor implements it as a bridge for message passing.

There is an example called "sandbox" showing how it works:

```rust
    let path = Path::new("programs/examples/hello_starknet.cairo");

    let contract = compile_path(
        path,
        None,
        CompilerConfig {
            replace_ids: true,
            ..Default::default()
        },
    )
    .unwrap();

    let path = std::env::current_dir()
        .unwrap()
        .join("target/debug/cairo-executor");
    let sandbox = IsolatedExecutor::new(&path).unwrap();

    let entry_point = contract.entry_points_by_type.external.get(0).unwrap();

    let mut handler = SyscallHandler;

    // run program blocks, so ideally each isolated executor is ran on its own thread.
    let result = sandbox
        .run_program(
            contract.extract_sierra_program().unwrap(),
            vec![1.into()],
            Some(u128::MAX),
            entry_point.function_idx,
            &mut handler,
        )
        .unwrap();
    println!("{:#?}", result);

    // run more programs etc
```

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


igaray:
Moved over to Starknet in Rust in https://github.com/lambdaclass/starknet_in_rust/pull/1209

PR #429, opened 2024-02-06
Implement span_from_tuple

edg-l:
# TITLE

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/array.rs
+++ b/src/libfuncs/array.rs
@@ -84,7 +84,9 @@ where
         ArrayConcreteLibfunc::SnapshotPopBack(info) => {
             build_snapshot_pop_back(context, registry, entry, location, helper, metadata, info)
         }
-        ArrayConcreteLibfunc::SpanFromTuple(_) => todo!(),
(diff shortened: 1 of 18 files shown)


src/libfuncs/array.rs
```diff
@@ -1177,13 +1179,199 @@ where
...
+{
+    // tuple to array span (t,t,t) -> &[t,t,t]
+
+    if metadata.get::<ReallocBindingsMeta>().is_none() {
+        metadata.insert(ReallocBindingsMeta::new(context, helper));
+    }
+
+    let struct_type_info = registry.get_type(&info.ty)?;
+
+    let struct_ty = registry.build_type(context, helper, registry, metadata, &info.ty)?;
```
azteca1998, line 1206: I think we should add a check to ensure that every struct member has the same type.
The Cairo compiler probably guarantees this, but I'd add an assert just in case.

azteca1998 (approved):

PR #434, opened 2024-02-13
Fix single-variant enum bug.

azteca1998:
# Fix single-variant enum bug

## Description

Single-variant enums generate a struct with an `i0` tag, which is apparently not valid in LLVM (but LLVM didn't throw an error when used through at least invoke_packed).

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


edg-l (dismissed):

juanbono (changes requested):
Please include a test

juanbono (approved):

PR #461, opened 2024-03-01
Refactor and fix array and span libfuncs

azteca1998:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


cairo-tests/src/dict_test.cairo
```diff
@@ -135,6 +135,7 @@ fn test_bool_dict() {
 }
 
 #[test]
+#[ignore]
```
edg-l, line 138: why are these tests being ignored? can you add a comment if they aren't working anymore?

azteca1998: They're failing. Yesterday I thought it was because of the felt to bool bug but today it doesn't make that much sense to me anymore. I'll check it again just in case.

edg-l (approved):

igaray (approved):

PR #463, opened 2024-03-05
Bump to 2.6-rc1 & Minor Bug Fix

rodrigo-pino:
We bumped the Cairo dependencies to 2.6-rc1 and also found a small bug.

The bug, found & fixed by @DomHenderson, is caused by the `build_slice` function in `src/libfuncs/array.rs`. It calculates the length to copy at runtime instead of statically.

A small fixed was to change from `llvm.memcpy.inline` to `llvm.memcpy` although it can also be fixed as well by calculating the length at compile time. The reason `llvm.memcpy.inline` is causing the error seems to be to the `.inline` variation which should only be used for immediate arguments and not variables (https://llvm.org/docs/LangRef.html#llvm-memcpy-inline-intrinsic , the argument sub-section).

As an example any code that uses `array_slice` libfunc should crash. I can provide one in a while if required

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -406,9 +406,9 @@ dependencies = [
 
 [[package]]
 name = "cairo-lang-casm"
-version = "2.5.4"
+version = "2.6.0-rc.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8ceb71a4cbf5b474bd671c79b2c05e8168a97199bfea1c01ef63b1bdaac3db03"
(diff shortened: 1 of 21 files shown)


rodrigo-pino:
Well, after a closer look, it seems the fix was already included, so this would be just a version bump 😅

edg-l:
Closing this as the main progress towards 2.6.3 is done in #504

PR #475, opened 2024-03-13
Improve error reporting

rodrigo-pino:
When returning an error, the error message was formed from only the first element of the segment. This updates the errors message to include all elements.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/execution_result.rs
+++ b/src/execution_result.rs
@@ -90,11 +90,14 @@ impl ContractExecutionResult {
                             })
                             .collect();
 
-                        let str_error =
-                            String::from_utf8(felt_vec.first().unwrap().to_bytes_be().to_vec())
-                                .unwrap()
-                                .trim_start_matches('\0')
-                                .to_owned();
+                        let bytes_err: Vec<_> = felt_vec
+                            .iter()
+                            .flat_map(|felt| felt.to_bytes_be().to_vec())
+                            // remove null chars
+                            .filter(|b| *b != 0)
+                            .collect();
+                        let str_error = String::from_utf8(bytes_err).unwrap().to_owned();
+
                         error_msg = Some(str_error);
                         felt_vec
                     } else {


src/execution_result.rs
```diff
@@ -90,11 +90,14 @@ impl ContractExecutionResult {
...
-                                .unwrap()
-                                .trim_start_matches('\0')
-                                .to_owned();
+                        let bytes_err: Vec<_> = felt_vec
+                            .iter()
+                            .flat_map(|felt| felt.to_bytes_be().to_vec())
+                            // remove null chars
+                            .filter(|b| *b != 0)
+                            .collect();
+                        let str_error = String::from_utf8(bytes_err).unwrap().to_owned();
```
azteca1998, line 99: Do you know if skipping null characters instead of treating them as an end-of-string is the correct behavior? My intuition tells me there's nothing of interest after a null byte but we're talking about felts, not string, so I could be wrong.

I'm curious as to why not change the `error_msg` field to `Vec<String>` instead of concatenating everything into a single string. This way it'd keep the original message structure.

An idea I've just had (for the future, not a request for this PR) is to make the `error_msg` field as its own struct (as in `error_msg: ErrorMessage`) that stores the `Vec<Felt>` and provides both the felt values and an implementation of `ToString` with the concatenation.

rodrigo-pino: > Do you know if skipping null characters instead of treating them as an end-of-string is the correct behavior? My intuition tells me there's nothing of interest after a null byte but we're talking about felts, not string, so I could be wrong.

Because each felt occupies 31 bytes instead of the usual 32, it means at the beginning, between Felt slots and at the end (if the full slot wasn’t used) they’ll be null chars. They won’t impact the string printing but they will exist on their structure which can produce unexpected results (e.g. longer length and affect string comparison)

> I'm curious as to why not change the error_msg field to Vec<String> instead of concatenating everything into a single string. This way it'd keep the original message structure.

Do you mean a Vec<string> where each element is a decoded felt? I personally don’t think it would be better since a word may be shared among different slots.  You’ll need to print each element of the Vec<String> together which is the same as being concatenated in the first place.

azteca1998: > Because each felt occupies 31 bytes instead of the usual 32, it means at the beginning, between Felt slots and at the end (if the full slot wasn’t used) they’ll be null chars. They won’t impact the string printing but they will exist on their structure which can produce unexpected results (e.g. longer length and affect string comparison)

I think I made you misunderstand the question by mistake, I forgot that in big endian the data is at the end. What I meant is the difference between the felt `'a\x00b\x00c'` being printed as:
  - `a\x00b\x00c` as in our current implementation, treating all intermediate null bytes as part of the string
  - `abc` as you're proposing

My rational is that if there are null bytes within the string then they're probably intentional, but it's not something I know for a fact to be the correct behavior, that's why I was asking.

> Do you mean a Vec where each element is a decoded felt? I personally don’t think it would be better since a word may be shared among different slots. You’ll need to print each element of the Vec together which is the same as being concatenated in the first place.

Makes sense, thanks for the explanation.

rodrigo-pino: I think I understand,  I hadn’t thought about an edge case where people would intentionally add null chars to their error messages (perhaps some secret protocol 😅).  This implementation is deleting both null chars, both created as a byproduct of the way the `Felt` type is modeled or intentionally written in. It is a small change, I'll push a fixed solution shortly. 

You also made me wonder about the possible usefulness of Null Chars and it turns out in modern languages such as rust they don't have any special feature, there are a remnant from past languages such as C & C++ where they were used to mark the end of a string, among other things.

azteca1998: In C you use arrays for both binary and text data, but C++ (when using the STL) and Rust make the distinction explicit by using `std::vector<uint8_t>`/`std::string` and `Vec<u8>`/`String` respectively.

In the microcontroller era, you could do something like `while ((ch = *ptr++) != 0) do_something(ch);` which would use less registers (and therefore be faster) than `for (int i = 0; i < len; i++) do_something(ptr[len]);` which needed the `i`. I find those kind of clever optimizations really interesting, but I agree that today they're a bit obsolete.

rodrigo-pino: After some thinking though, I think there is no perfect solution. It can be either:

1. purge all null bytes
2. leave some (determined by some rules) how do you determine which one is the complicated part because it depends on how a user uses the encoding
3. leave them all

I am personally Ok with (1) & (3). (2) can make my life easier or not depending on the rules. 

If it were up to me, I'll will purge them all. There a lot of Chars you can use a special char in case of a custom protocol. Also, the null chars are only being removed from the string message, the Return segment remains the same.  Finally, I find improbable for anyone to use the error message for something else than printing/testing

azteca1998: I was experimenting with the Cairo VM and I've found something interesting:
```cairo
```
This program panic!()s with:
```
Run panicked with [172063216033151516844329818169388221396727601204421676283161692175877681972 ('abcdefghijklmnopqrstuvwxyz01234'), 58506970937648 ('567890'), 106654272086784, ].
```
Seeing this I'd say that the best way to do it would be to keep the `Vec<Felt>`, maybe wrapped in a struct that provides a method to extract the strings; but since that's a bit more complicated and outside the scope of this PR I guess it's fine as it is.

azteca1998 (approved):

src/execution_result.rs
rodrigo-pino: It's late here, I am not understanding fully what's actually happening with the last value. I'll take a look at it tomorrow

edg-l (approved):

src/execution_result.rs
rodrigo-pino: After thinking about it, I personally think that the best solution would be to remove the existence of an error message in the `ExecutionResult` type. Let each tool that uses  `ExecutionResult` decide how they will interpret  the Vec<Felt> as they see fit depending on their requirements. Any interpretation given by default will be slightly opinionated. Also, I am just a passerby that don't have the whole picture so feel free to dismiss it.

azteca1998: > It's late here, I am not understanding fully what's actually happening with the last value. I'll take a look at it tomorrow

The last value contains non-printable characters and is therefore interpreted as not a string.

> After thinking about it, I personally think that the best solution would be to remove the existence of an error message in the ExecutionResult type.

Yes, not necessarily remove it but keep the original data so that the user can interpret it however they need to. Right now, however, it's fine as it is. Thanks for your PR.

PR #489, opened 2024-03-19
PR: Improve the error handling code

### LINKED ISSUE #488: Improve the error handling code
We currently have four implementations of error types in https://github.com/lambdaclass/cairo_native/tree/main/src/error . 
These four error types also overlap somewhat. It would be nicer to have a single `Error` enum type, gathering all error cases into separate variantes, and also removing the `ErrorImpl`.

hhamud:
* fixes #488

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added


hhamud:
@igaray can you review?

src/error.rs
```diff
@@ -1,11 +1,118 @@
...
+use std::{alloc::LayoutError, fmt, num::TryFromIntError, ops::Deref};
+use thiserror::Error;
+
+use crate::metadata::gas::GasMetadataError;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, Error)]
+pub struct Error {
+    pub source: Box<ErrorImpl>,
```
edg-l, line 15: i think Error itself can be the ErrorImpl, and if possible remove the box

azteca1998: I've just checked it and it seems `std::backtrace::Backtrace` is stable since Rust 1.65.0.
If you're going to add the error location backtraces please add them in `Error`, otherwise there's no point to having `Error` wrap `ErrorImpl` as @edg-l said.

hhamud: I tried to do this but the same errors pop up and Dtolnay says that its still not ready to be used with his package. So I've gone and removed the errorimp instead. I am on rust 1.76.

edg-l (dismissed):

azteca1998 (dismissed):

azteca1998 (approved):

Oppen (approved):

igaray (approved):

PR #518, opened 2024-04-17
Refactor common operations into a block trait

edg-l:
We have a lot of repeated places in codegen where we use complicated or repetitive operations, this refactors some common complexity so we can avoid mistakes.

For now I only use the methods in a few places like the array libfuncs, but it could be used in all libfuncs, It's just quite some more work to do replacing those.

Operations that are in BlockExt that can be replaced

- llvm.alloca
- arith.constant
- llvm.extract_value
- llvm.insert_value
- llvm.alloca
- llvm.load
- llvm.store
- llvm.memcpy

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/block_ext.rs
+++ b/src/block_ext.rs
@@ -317,9 +317,7 @@ impl<'ctx> BlockExt<'ctx> for Block<'ctx> {
             ));
         }
 
-        self.append_op_result(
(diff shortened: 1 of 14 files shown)


igaray (approved):

elielnfinic:
Can I take this?

PR #534, opened 2024-04-18
Add unit tests for `libfuncs` `switch` function.

### LINKED ISSUE #484: Test the switch libfunc helper function
The `switch` function in `src/libfuncs.rs` is a helper function used to implement libfuncs and is not directly tested, although the code generated is tested in other places. It would be good to find a way to test it directly and increase [coverage](https://app.codecov.io/gh/lambdaclass/cairo_native/blob/main/src%2Flibfuncs.rs) of the libfuncs module.

tcoratger:
Add unit tests for `libfuncs` `switch` function in order to test various configurations.

Resolves #484.

## Checklist
- [X] Linked to Github Issue
- [X] Unit tests added

### DIFF
--- a/src/libfuncs.rs
+++ b/src/libfuncs.rs
@@ -511,28 +511,14 @@ pub fn increment_builtin_counter<'ctx: 'a, 'a>(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use melior::dialect::DialectRegistry;
+    use crate::context::NativeContext;
     use melior::ir::Type;
-    use melior::utility::register_all_dialects;
-
-    /// Load all available MLIR dialects into the given context.
-    ///
-    /// This function initializes a new `DialectRegistry`, registers all available dialects
-    /// into it, appends the registry to the provided context, and then loads all available
-    /// dialects into the context.
-    pub fn load_all_dialects(context: &Context) {
-        let registry = DialectRegistry::new();
-        register_all_dialects(&registry);
-        context.append_dialect_registry(&registry);
-        context.load_all_available_dialects();
-    }
 
     #[test]
     fn switch_branch_arg_external_test() {
         // Create a new context for MLIR operations
-        let context = Context::new();
-        // Load all available MLIR dialects into the context
-        load_all_dialects(&context);
+        let native_context = NativeContext::new();
+        let context = native_context.context();
 
         // Create an unknown location in the context
         let location = Location::unknown(&context);
@@ -606,9 +592,8 @@ mod tests {
     #[test]
     fn switch_branch_arg_returned_test() {
         // Create a new context for MLIR operations
-        let context = Context::new();
-        // Load all available MLIR dialects into the context
-        load_all_dialects(&context);
+        let native_context = NativeContext::new();
+        let context = native_context.context();
 
         // Create an unknown location in the context
         let location = Location::unknown(&context);
@@ -691,9 +676,8 @@ mod tests {
     #[test]
     fn switch_branch_target_jump_test() {
         // Create a new context for MLIR operations
-        let context = Context::new();
-        // Load all available MLIR dialects into the context
-        load_all_dialects(&context);
+        let native_context = NativeContext::new();
+        let context = native_context.context();
 
         // Create an unknown location in the context
         let location = Location::unknown(&context);
--- a/src/libfuncs.rs
+++ b/src/libfuncs.rs
@@ -543,10 +543,10 @@ mod tests {
         // Create an integer type with 32 bits
         let i32_type: Type = IntegerType::new(&context, 32).into();
         // Create a default block with the integer type and the unknown location
-        let default_block = Block::new(&[(i32_type, location)]);
+        let default_block = lib_func_helper.append_block(Block::new(&[(i32_type, location)]));
 
         // Create a new MLIR block
-        let block = Block::new(&[]);
+        let block = lib_func_helper.append_block(Block::new(&[]));
 
         // Append a constant arithmetic operation to the block and obtain its result operand
         let operand = block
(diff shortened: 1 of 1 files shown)


src/libfuncs.rs
```diff
@@ -507,3 +507,258 @@ pub fn increment_builtin_counter<'ctx: 'a, 'a>(
...
+    use melior::dialect::DialectRegistry;
+    use melior::ir::Type;
+    use melior::utility::register_all_dialects;
+
+    /// Load all available MLIR dialects into the given context.
+    ///
+    /// This function initializes a new `DialectRegistry`, registers all available dialects
+    /// into it, appends the registry to the provided context, and then loads all available
+    /// dialects into the context.
+    pub fn load_all_dialects(context: &Context) {
```
edg-l, line 523: Please use the NativeContext struct, it handles all of this, the dialect loading, etc.

src/libfuncs.rs
```diff
@@ -507,3 +507,258 @@ pub fn increment_builtin_counter<'ctx: 'a, 'a>(
...
+        // Load all available MLIR dialects into the context
+        load_all_dialects(&context);
+
+        // Create an unknown location in the context
+        let location = Location::unknown(&context);
+        // Create a new MLIR module with the unknown location
+        let module = Module::new(location);
+
+        // Create a new MLIR block and obtain its reference
+        let last_block = unsafe { &BlockRef::from_raw(Block::new(&[]).into_raw()) };
```
edg-l, line 543: You dont need unsafe to get a BlockRef.

Also first you need to create a function with the `func` dialect, check out the melior crate example. There you will need to add a region, and you add blocks to the region, when adding the block to the region you will get a BlockRef

edg-l (changes requested):

src/libfuncs.rs
tcoratger: Fixed

src/libfuncs.rs
tcoratger: Nice, just fixed. As I can have an empty block here, I didn't add a function inside it as you pointed out inside your comment. I think it's useless in this situation, but I may be wrong. Let me know if it's ok for you.

src/libfuncs.rs
```diff
@@ -507,3 +507,248 @@ pub fn increment_builtin_counter<'ctx: 'a, 'a>(
...
+
+        // Create a new MLIR block and obtain its reference
+        let block = Block::new(&[]);
+        let region = Region::new();
+        let last_block = &region.append_block(block);
+
+        // Initialize the LibfuncHelper struct with various parameters
+        let mut lib_func_helper = LibfuncHelper {
+            module: &module,
+            init_block: unsafe { &BlockRef::from_raw(Block::new(&[]).into_raw()) },
```
edg-l, line 536: there is still this unsafe which is not needed.

As a general rule, we usually dont hold a Block, we add blocks using the region like this:

```rust
let last_block = &region.append_block(Block::new(...));
```

or the libfunc helper:

```
helper.append_block(Block::new(...))
```

edg-l (changes requested):
The unsafe is still there, i leave you here how we build the libfunc helper in the compiler: 

https://github.com/lambdaclass/cairo_native/blob/ca6549a68c1b4266a7f9ea41dc196bf4433a2ee8/src/compiler.rs#L342-L360

src/libfuncs.rs
edg-l: You are right about not needing the function.

src/libfuncs.rs
tcoratger: Sorry didn't notice this one. Can you confirm this is ok now?

src/libfuncs.rs
```diff
@@ -507,3 +507,245 @@ pub fn increment_builtin_counter<'ctx: 'a, 'a>(
...
+            blocks_arena: &Bump::new(),
+            last_block: Cell::new(&last_block),
+            branches: Vec::new(),
+            results: Vec::new(),
+        };
+
+        // Create an integer type with 32 bits
+        let i32_type: Type = IntegerType::new(&context, 32).into();
+        // Create a default block with the integer type and the unknown location
+        let default_block = Block::new(&[(i32_type, location)]);
```
edg-l, line 546: use `lib_func_helper.append_block` on this block too.

src/libfuncs.rs
```diff
@@ -507,3 +507,245 @@ pub fn increment_builtin_counter<'ctx: 'a, 'a>(
...
+
+        // Create a new MLIR block
+        let block = Block::new(&[]);
```
edg-l, line 549: same here, `lib_func_helper.append_block(Block::new(...))`

src/libfuncs.rs
```diff
@@ -507,3 +507,245 @@ pub fn increment_builtin_counter<'ctx: 'a, 'a>(
...
+            blocks_arena: &Bump::new(),
+            last_block: Cell::new(&last_block),
+            branches: Vec::new(),
+            results: Vec::new(),
+        };
+
+        // Create an integer type with 32 bits
+        let i32_type: Type = IntegerType::new(&context, 32).into();
+        // Create a default block with the integer type and the unknown location
+        let default_block = Block::new(&[(i32_type, location)]);
```
edg-l, line 622: same as the other test

edg-l (changes requested):
Make sure all blocks are in the region, you can do so with `lib_func_helper.append_block` when doing `Block::new`

src/libfuncs.rs
tcoratger: fixed

src/libfuncs.rs
tcoratger: fixed

src/libfuncs.rs
tcoratger: fixed

edg-l (dismissed):

tcoratger:
@edg-l Is there anything wrong with the ci?

edg-l (approved):

igaray (approved):

Issue #526, opened 2024-04-18
Refactor enum libfuncs to use the BlockExt trait


edg-l:
See: https://github.com/lambdaclass/cairo_native/pull/518

Refactor the code implementing the libfuncs to use the trait methods from `BlockExt`.

PabloVillaplana:
Hello @edg-l I would love to be part of this effort! 👷🏽‍♂️, thanks (I am in OD Hack telegram, ready to start on Monday 🫡)

g4titanx:
hello @edg-l , i would love to work on this

edg-l:
> hello @edg-l , i would love to work on this

Sure thing!

PR #558, opened 2024-04-30
Refactored enum libfuncs to use the BlockExt trait

g4titanx:
## Checklist
- [ ] Linked to Github Issue: https://github.com/lambdaclass/cairo_native/issues/526
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/enum.rs
+++ b/src/libfuncs/enum.rs
@@ -95,7 +95,7 @@ pub fn build_init<'ctx, 'this>(
                 false,
             );
 
-            let tag_val = entry.const_int(context, location, info.index as i64, 64)?;
+            let tag_val = entry.const_int_from_type(context, location, tag_ty, info.index)?;
 
             let payload_val = if payload_type_info.is_memory_allocated(registry) {
                 entry.load(context, location, entry.argument(0)?.into(), variant_tys[info.index].0, Some(variant_tys[info.index].1.align()))?
@@ -124,23 +124,14 @@ pub fn build_init<'ctx, 'this>(
             };
 
             if type_info.is_memory_allocated(registry) {
-                let k1: i64 = helper
-                    .init_block()
-                    .append_operation(arith::constant(
-                        context,
-                        IntegerAttribute::new(IntegerType::new(context, 64).into(), 1).into(),
-                        location,
-                    ))
-                    .result(0)?
-                    .into();
-            let stack_ptr = helper.init_block().alloca1(
-                context,
-                location,
-                type_info.build(
+                let stack_ptr = helper.init_block().alloca1(
                     context,
-                    helper,
-                    registry,
-                    metadata,
+                    location,
+                    type_info.build(
+                        context,
+                        helper,
+                        registry,
+                        metadata,
                     &info.branch_signatures()[0].vars[0].ty,
                 )?,
                 Some(layout.align()),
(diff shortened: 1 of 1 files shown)


src/libfuncs/enum.rs
```diff
@@ -359,33 +281,9 @@ pub fn build_match<'ctx, 'this>(
...
-                            AllocaOptions::new()
-                                .align(Some(IntegerAttribute::new(
-                                    IntegerType::new(context, 64).into(),
-                                    payload_layout.align() as i64,
-                                )))
-                                .elem_type(Some(TypeAttribute::new(payload_ty))),
-                        ))
-                        .result(0)?
-                        .into();
+                    let stack_ptr = helper.alloca(context, location, k1, Some(payload_layout.align()))?;
```
edg-l, line 284: should be `helper.init_block()` here.

Also you can use `alloca1` to avoid creating the k1 constant.

I also think you are missing some arguments there.

src/libfuncs/enum.rs
```diff
@@ -160,39 +133,9 @@ pub fn build_init<'ctx, 'this>(
...
-                                context,
-                                helper,
-                                registry,
-                                metadata,
-                                &info.branch_signatures()[0].vars[0].ty,
-                            )?))),
-                    ))
-                    .result(0)?
-                    .into();
+            let stack_ptr = helper.alloca(context, location, k1, Some(layout.align()))?;
```
edg-l, line 136: look below

edg-l (changes requested):

src/libfuncs/enum.rs
g4titanx: oh, adding the operations to init block?

azteca1998: Yes, please. Otherwise the code is not equivalent to the original.

src/libfuncs/enum.rs
g4titanx: fixed

src/libfuncs/enum.rs
g4titanx: right! fixed.

src/libfuncs/enum.rs
```diff
@@ -94,29 +95,10 @@ pub fn build_init<'ctx, 'this>(
...
 
-            let tag_val = entry
-                .append_operation(arith::constant(
-                    context,
-                    IntegerAttribute::new(tag_ty, info.index as i64).into(),
-                    location,
-                ))
-                .result(0)?
-                .into();
+            let tag_val = entry.const_int(context, location, info.index as i64, 64)?;
```
edg-l, line 98: the cast is not needed here, also you should use the tag_ty, so instead of const_int use const_int_from_type and pass tag_ty along the info.index

src/libfuncs/enum.rs
```diff
@@ -151,7 +124,7 @@ pub fn build_init<'ctx, 'this>(
             };
 
             if type_info.is_memory_allocated(registry) {
-                let k1 = helper
+                let k1: i64 = helper
```
edg-l, line 127: k1 is probably unused and can be removed if we use alloca1.

src/libfuncs/enum.rs
```diff
@@ -359,33 +292,9 @@ pub fn build_match<'ctx, 'this>(
...
-                            AllocaOptions::new()
-                                .align(Some(IntegerAttribute::new(
-                                    IntegerType::new(context, 64).into(),
-                                    payload_layout.align() as i64,
-                                )))
-                                .elem_type(Some(TypeAttribute::new(payload_ty))),
-                        ))
-                        .result(0)?
-                        .into();
+                    let stack_ptr = helper.init_block().alloca1(context, location, payload_layout, Some(payload_layout.align()))?;
```
edg-l, line 295: i think the first payload_layout here should be payload_ty

edg-l (changes requested):

src/libfuncs/enum.rs
g4titanx: oh, okay, changed payload_layout to payload_ty

juanbono (changes requested):
Conflict in code and several CI steps are broken

Issue #468, opened 2024-03-12
Bug: cairo test test_ecdsa_recover_public_key fails


edg-l:
Fails with recover_ecdsa_public_key failed

```cairo
// ./cairo-tests/src/ec_test.cairo

// TODO: fails  recover_ecdsa_public_key failed
#[test]
#[ignore]
fn test_ecdsa_recover_public_key() {
    let message_hash = 0x503f4bea29baee10b22a7f10bdc82dda071c977c1f25b8f3973d34e6b03b2c;
    let signature_r = 0xbe96d72eb4f94078192c2e84d5230cde2a70f4b45c8797e2c907acff5060bb;
    let signature_s = 0x677ae6bba6daf00d2631fab14c8acf24be6579f9d9e98f67aa7f2770e57a1f5;
    assert_eq(
        @ecdsa::recover_public_key(:message_hash, :signature_r, :signature_s, y_parity: false)
            .unwrap(),
        @0x7b7454acbe7845da996377f85eb0892044d75ae95d04d3325a391951f35d2ec,
        'recover_ecdsa_public_key failed'
    );
    assert(
        ecdsa::check_ecdsa_signature(
            :message_hash,
            public_key: ecdsa::recover_public_key(
                :message_hash, :signature_r, :signature_s, y_parity: true
            )
                .unwrap(),
            :signature_r,
            :signature_s
        ),
        'ecdsa returned false'
    );
}
```

muheebyusufbaba1:
Hello, I love to work on this

igaray:
Go for it!

xiaolou86:
Hello @igaray @edg-l , I want to work on this.

PR #569, opened 2024-05-05
Fix `test_ecdsa_recover_public_key` test

tcoratger:
Should close https://github.com/lambdaclass/cairo_native/issues/468

The modular inverse should be obtained roughly via the pseudo code:

```
function inverse(a, n)
    t := 0;     newt := 1
    r := n;     newr := a

    while newr ≠ 0 do
        quotient := r div newr
        (t, newt) := (newt, t − quotient × newt) 
        (r, newr) := (newr, r − quotient × newr)

    if r > 1 then
        return "a is not invertible"
    if t < 0 then
        t := t + n

    return t
```

In the `build_u256_guarantee_inv_mod_n` construction, we missed the `t < 0` condition, causing this operation https://github.com/starkware-libs/cairo/blob/afeeed29b37bd394535fcf3e8d7d16db047381fb/corelib/src/ecdsa.cairo#L143 to fail in the ECDSA `recover_public_key` procedure. This issue is fixed and another unit test is added in the `u256_inv_mod_n` suite to cover this scenario.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/cairo-tests/src/ec_test.cairo
+++ b/cairo-tests/src/ec_test.cairo
@@ -111,9 +111,7 @@ fn test_ecdsa() {
     );
 }
 
-// TODO: fails  recover_ecdsa_public_key failed
 #[test]
-#[ignore]
 fn test_ecdsa_recover_public_key() {
     let message_hash = 0x503f4bea29baee10b22a7f10bdc82dda071c977c1f25b8f3973d34e6b03b2c;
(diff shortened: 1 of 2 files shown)


edg-l (approved):

edg-l:
Nice find! Thanks a lot.

azteca1998 (approved):
Nice work, thanks for the fix.

PR #588, opened 2024-05-13
Add Cheatcode Syscalls

### LINKED ISSUE #103: Libfuncs: starknet testing
[Sierra implementation](https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra/src/extensions/modules/starknet/testing.rs)
  - [ ] set_block_number
  - [ ] set_block_timestamp
  - [ ] set_caller_address
  - [ ] set_contract_address
  - [ ] set_sequencer_address

juanbono:
Implements the cheatcode libfunc (a syscall)

closes #103 

Adds a new method to the SyscallHandler trait: `cheatcode` in `src/starknet.rs`.

Implements the compilation of the cheatcode libfunc in `src/libfuncs/starknet/testing.rs`. Usually this involves accessing the SyscallHandlerCallbacks through it's arguments and calling the appropiate method, but Sierra doesn't specify that this particular syscall needs the syscall handler, so we don't receive the SyscallHandlerCallbacks as an argument.

The workaround is to, at the start of the execution (`src/executor.rs`), store a pointer to the SyscallHandlerCallbacks in a thread LocalKey and access it with a runtime function defined in (`src/starknet.rs`) that accesses the SyscallHandler and executes the cheatcode syscall. This runtime function is called from MLIR.

Adds a dummy implementation of the syscall in `examples/starknet.rs` and `tests/tests/starknet/syscalls.rs` and tests some of them. The behaviour of most syscalls is the same so not all of them were tested to avoid unnecesary repetition.

## Change Summary

A summary of changes in order of importance:

- `src/libfuncs/starknet/testing.rs`: Implements libfunc compilation
- `src/executor.rs`: Set thread LocalKey pointer to SyscallHandlerCallbacks
- `src/starknet.rs`: Add cheatcode method to trait and runtime cheatcode wrapper
- `src/metadata/runtime_bindings.rs`: Add  helper function to call runtime cheatcode wrapper
- `tests/tests/starknet/syscalls.rs`: Test cheatcode syscall
- `examples/starknet.rs`: Add cheatcode implementation to SyscallHandler

## Cheatcodes

- [x] pop_log
- [x] pop_l2_to_l1_message
- [x] set_account_contract_address
- [x] set_block_number
- [x] set_block_timestamp
- [x] set_caller_address
- [x] set_chain_id
- [x] set_contract_address
- [x] set_max_fee
- [x] set_nonce
- [x] set_sequencer_address
- [x] set_signature
- [x] set_transaction_hash
- [x] set_version

## Checklist

- [x] Linked to Github Issue
- [x] Integration tests added


src/starknet.rs
```diff
@@ -584,10 +542,14 @@ pub(crate) mod handler {
...
 
+    /// A C ABI Wrapper around the StarknetSyscallHandler
+    ///
+    /// It contains pointers to functions which can be called through MLIR based on the field offset
+    /// The functions convert C ABI structures to the Rust equivalent and calls the wrapped implementation.
     #[repr(C)]
     #[derive(Debug)]
     pub struct StarknetSyscallHandlerCallbacks<'a, T> {
-        self_ptr: &'a mut T,
+        pub self_ptr: &'a mut T,
```
juanbono, line 552: is there a reason for making this public?

JulianGCalderon: Originally I was accessing `self_ptr` through `StarknetSyscallHandlerCallbacks` instead of dereferencing the pointer. Now there is no need to have it public, i'm going to correct it.

JulianGCalderon: Done

src/starknet.rs
```diff
@@ -481,6 +435,10 @@ impl StarknetSyscallHandler for DummySyscallHandler {
     ) -> SyscallResult<(U256, U256)> {
         unimplemented!()
     }
+
+    fn cheatcode(&mut self, _selector: Felt, _input: &[Felt]) -> Vec<Felt> {
+        todo!()
+    }
```
azteca1998, line 441: Same as before.

src/starknet.rs
```diff
@@ -584,6 +542,10 @@ pub(crate) mod handler {
         nonce: Felt252Abi,
     }
 
+    /// A C ABI Wrapper around the StarknetSyscallHandler
+    ///
+    /// It contains pointers to functions which can be called through MLIR based on the field offset
+    /// The functions convert C ABI structures to the Rust equivalent and calls the wrapped implementation.
```
azteca1998, line 548: Thanks for the docs. Since I've been asked that before, could you add that the reason for having to have the function pointers instead of it being like the runtime library is that the vtable has a generic type?

src/starknet.rs
```diff
@@ -1630,3 +1627,37 @@ pub(crate) mod handler {
         }
     }
 }
+
+thread_local!(pub static SYSCALL_HANDLER_VTABLE: std::cell::Cell<*mut ()>  = const { std::cell::Cell::new(null_mut()) });
```
azteca1998, line 1631: There`s two space characters between the type and the assignment symbol (`=`).
Is the const block necessary? And the Cell too, I've seen that `LocalKey` has methods to mutate it but I can't seem to make an example work in the rust playground.

src/utils.rs
```diff
@@ -1434,5 +1439,9 @@ pub mod test {
         ) -> SyscallResult<(U256, U256)> {
             unimplemented!()
         }
+
+        fn cheatcode(&mut self, _selector: Felt, _input: &[Felt]) -> Vec<Felt> {
+            todo!()
+        }
```
azteca1998, line 1445: Same as before.

tests/tests/starknet/keccak.rs
```diff
@@ -262,6 +262,10 @@ impl StarknetSyscallHandler for SyscallHandler {
     ) -> SyscallResult<(U256, U256)> {
         unimplemented!()
     }
+
+    fn cheatcode(&mut self, _selector: Felt, _input: &[Felt]) -> Vec<Felt> {
+        todo!()
+    }
```
azteca1998, line 268: Same as before.

tests/tests/starknet/secp256.rs
```diff
@@ -247,6 +247,10 @@ impl StarknetSyscallHandler for &mut SyscallHandler {
         args.push_back(p);
         Ok(rets.pop_front().unwrap())
     }
+
+    fn cheatcode(&mut self, _selector: Felt, _input: &[Felt]) -> Vec<Felt> {
+        todo!()
+    }
```
azteca1998, line 253: Same as before.

src/starknet.rs
```diff
@@ -1630,3 +1627,37 @@ pub(crate) mod handler {
...
+    let ptr = SYSCALL_HANDLER_VTABLE.with(|ptr| ptr.get());
+    if ptr.is_null() {
+        *result_ptr = ArrayAbi {
+            ptr: null_mut(),
+            since: 0,
+            until: 0,
+            capacity: 0,
+        };
+        return;
+    }
```
azteca1998, line 1652: Why not an assert? The way I see it, if the cheatcode is called and there's not a valid syscall handler there's not really much to do other than crash.

azteca1998 (commented):
I haven't seen the mechanism to make cheatcodes work properly on multilevel contract calls. For example:
  1. Contract A is invoked.
  2. Contract A's cheatcode handler is configured.
  3. Contract A invokes contract B.
  4. Contract B's cheatcode handler is configured. It overwrites contract A's handler.
  5. Contract B returns control to contract A.
  6. Contract A invokes a cheatcode, potentially crashing the program because the cheatcode handler is not its own.

To fix this you should:
  - Store the original value of `SYSCALL_HANDLER_VTABLE`.
  - Set the correct cheatcode handler at `invoke_dynamic` in `src/executor.rs` (you're already doing it).
  - Restore the original value when returning from `invoke_dynamic` in `src/executor.rs`.

If my reasoning is correct this should cover every case, but just in case if you discover a suspicious control flow that could crash using those directives please tell me and I'll check it.

src/starknet.rs
JulianGCalderon: You're right. I'm going to change it to an assert.

src/starknet.rs
JulianGCalderon: Nice catch on the duplicated space. Probably `cargo fmt` didn't format it as it's inside a macro

The const block was suggested by clippy: `warning: initializer for thread_local value can be made const`

I think the cell is necessary, `replace` is only implemented for LocalKey<Cell<T>>: https://doc.rust-lang.org/std/thread/struct.LocalKey.html#impl-LocalKey%3CCell%3CT%3E%3E

examples/erc20.rs
```diff
@@ -265,6 +265,10 @@ impl StarknetSyscallHandler for SyscallHandler {
     ) -> SyscallResult<(U256, U256)> {
         unimplemented!()
     }
+
+    fn cheatcode(&mut self, _selector: Felt, _input: &[Felt]) -> Vec<Felt> {
+        todo!()
+    }
```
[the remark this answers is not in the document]
JulianGCalderon: Ahhh I see, it makes sense. I'm going to fix it.

src/executor.rs
```diff
@@ -196,6 +197,18 @@ fn invoke_dynamic(
...
+    // so we must always allocate it in case it needs it, regardless of whether it's passed
+    // as an argument to the entry point or not.
+    let mut syscall_handler = syscall_handler.as_mut().map(|syscall_handler| {
+        let syscall_handler = arena.alloc(StarknetSyscallHandlerCallbacks::new(syscall_handler));
+
+        let syscall_handler_ptr = std::ptr::addr_of!(*syscall_handler) as *mut ();
+        SYSCALL_HANDLER_VTABLE.set(syscall_handler_ptr);
+
+        syscall_handler
+    });
```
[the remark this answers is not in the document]
JulianGCalderon: Aaaah, I see. I'm going to try it that way. Thanks for the suggestions.

examples/erc20.rs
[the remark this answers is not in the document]
JulianGCalderon: Done

src/starknet.rs
JulianGCalderon: Done (removed the duplicated space)

src/starknet.rs
JulianGCalderon: Done

tests/tests/starknet/secp256.rs
JulianGCalderon: Done

tests/tests/starknet/keccak.rs
JulianGCalderon: Done

src/utils.rs
JulianGCalderon: Done

src/bin/utils/test.rs
```diff
@@ -850,6 +850,10 @@ impl StarknetSyscallHandler for TestSyscallHandler {
     ) -> SyscallResult<(U256, U256)> {
         Ok((p.x, p.y))
     }
+
+    fn cheatcode(&mut self, _selector: Felt, _input: &[Felt]) -> Vec<Felt> {
+        todo!()
+    }
```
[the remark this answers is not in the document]
JulianGCalderon: Done

src/executor.rs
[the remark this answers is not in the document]
JulianGCalderon: Done. I also reset the ptr as you mentioned in the main comment.

src/executor.rs
```diff
@@ -252,6 +262,11 @@ fn invoke_dynamic(
         );
     }
 
+    // If the syscall handler was changed, then reset the previous one.
+    if let Some(previous_syscall_handler) = previous_syscall_handler {
+        SYSCALL_HANDLER_VTABLE.set(previous_syscall_handler);
+    }
```
azteca1998, line 268:
```suggestion
    if let Some(previous_syscall_handler) = previous_syscall_handler {
        SYSCALL_HANDLER_VTABLE.set(previous_syscall_handler);
    } else {
        SYSCALL_HANDLER_VTABLE.set(null_mut());
    }
```

azteca1998 (approved):

src/executor.rs
azteca1998: Nevermind it's ok as it is now, I got confused.

azteca1998 (dismissed):

pefontana (changes requested):
Can we move this feature under a feature flag?

JulianGCalderon:
Done @pefontana

azteca1998 (approved):

src/libfuncs/starknet.rs
```diff
@@ -138,7 +141,14 @@ pub fn build<'ctx, 'this>(
         StarkNetConcreteLibfunc::Secp256(selector) => self::secp256::build(
             context, registry, entry, location, helper, metadata, selector,
         ),
-        StarkNetConcreteLibfunc::Testing(_) => todo!("implement starknet testing libfunc"),
+        #[cfg(feature = "with-cheatcode")]
+        StarkNetConcreteLibfunc::Testing(TestingConcreteLibfunc::Cheatcode(info)) => {
```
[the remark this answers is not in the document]
JulianGCalderon: I think the empty line was rendered by Github, there isn't actually an empty line there.

pefontana (approved):

PR #603, opened 2024-05-16
Fix wrong upcast from unsigned to signed

edg-l:
It used a sign extend when it should use a zero extend. The bug was that we checked whether to use sext if either src or dst type are signed when we should only care about src type

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


juanbono (approved):

juanbono (approved):

juanbono:
Nice catch!

PR #610, opened 2024-05-17
Add Bytes31 return type

pefontana:
Add Bytes31 return type

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added


edg-l (approved):

src/executor.rs
```diff
@@ -654,6 +654,20 @@ fn parse_result(
                 value
             }
         },
+        CoreTypeConcrete::Bytes31(_) => match return_ptr {
+            Some(return_ptr) => JitValue::from_jit(return_ptr, type_id, registry),
+            None => {
+                #[cfg(target_arch = "x86_64")]
+                let value = JitValue::from_jit(return_ptr.unwrap(), type_id, registry);
```
azteca1998, line 661: I just realized that this unwrap here would always fail if it wasn't unreachable.
It would always fail because the match's branch means that it's always `None`, and it's unreachable because x86 only returns values in two registers at most (128 bits, not 248).

Maybe we should change it to `unreachable!()`? (here and in the felt)

azteca1998 (approved):

src/executor.rs
pefontana: You are right @azteca1998 !
I leave it like this 
```
        CoreTypeConcrete::Bytes31(_) => match return_ptr {
            // target_arch = "x86_64")]
            Some(return_ptr) => JitValue::from_jit(return_ptr, type_id, registry),
            None => {
                // target_arch = "aarch64"
                JitValue::Bytes31(unsafe {
                    *std::mem::transmute::<&[u64; 4], &[u8; 31]>(&ret_registers)
                })
            }
        },

```
Maybe in the feature we can refactor this function so it returns an error
What do you think?

pefontana: Nice catch!
I added unreachable!() for Bytes31 and Felt types. But I think we should modify the parse_result fn to return a Result, what do you think?

pefontana: Good catch! @azteca1998 
I added the unreachable!(), but I think I would be good to make a refactor in another PR that changes the fn parse_result so it return a Result, what do you think?

src/executor.rs
```diff
@@ -647,15 +647,24 @@ fn parse_result(
             Some(return_ptr) => JitValue::from_jit(return_ptr, type_id, registry),
             None => {
                 #[cfg(target_arch = "x86_64")]
-                let value = JitValue::from_jit(return_ptr.unwrap(), type_id, registry);
+                unreachable!();
```
igaray, line 650: Can you please add a comment explaining the difference behaviour across architectures?

igaray (changes requested):

src/executor.rs
azteca1998: Returning a result in the case of reaching an unreachable part of the code? I've never seen the point of doing that.
However, if it's a refactor that avoids the need to have an `unreachable!()` statement due to a change in the design (and not pushing it into a result) then I'm all for it.

src/executor.rs
azteca1998: Something like this should work:
```suggestion
                // Since x86_64's return values hold at most two different 64bit registers,
                // everything bigger than u128 will be returned by memory, therefore making
                // this branch unreachable on that architecture.
                unreachable!();
```

pefontana: Great!
Done [9d9b7a8](https://github.com/lambdaclass/cairo_native/pull/610/commits/9d9b7a8c8da573d23f8aec9a35f75191bebece52)

src/executor.rs
pefontana: Yes, I meant the refactor that avoids unreachable!()
I will take it next!

azteca1998 (approved):

juanbono (approved):

PR #612, opened 2024-05-20
small refactoring around `match`

tcoratger:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/executor.rs
```diff
@@ -802,24 +802,10 @@ fn parse_result(
                 registry,
             ),
         },
-        CoreTypeConcrete::Felt252DictEntry(_) => todo!(),
```
igaray, line 805: I know the variants are redundant with the catch-all branch below, but I feel that removing the explicit variants hides information that should be explicit, I would be more in favor of removing the catch-all if possible, or leaving as-is.

igaray (changes requested):

igaray (approved):

juanbono (approved):

PR #616, opened 2024-05-21
increase test coverage of metadata `DebugUtils`

tcoratger:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


Oppen:
Two things:
1. Most of the comments in the tests are superfluous, I'd rather keep the tests compact unless the comments actually clarify something;
2. The tests seem to test code generation is successful, but not the behavior of that code.

What I would expect from these tests is this:
1. They do test code generation.
2. They test the generated code behaves as expected (that is, a `print` writes to `stdout`, maybe check if you can override it with a buffer).
3. The `active_map` actually fits its purpose: we need to see both that:
  - It doesn't have the function marked as active before calling code generation.
  - The symbol is not emitted if we don't call that function.
  - It marks the function as active afterwards.
  - Successive calls don't emit the symbols twice.

azteca1998 (dismissed):
This module is for debugging, and it's not intended to even be compiled in normal builds so there's no need to test it more than we already do every time we need to use it. That said, any extra testing is appreciated.

edg-l:
The debug utils module is only used when debugging bugs internally in cairo native, as @azteca1998 pointed out I also don't think the tests are necesary for now

PR #618, opened 2024-05-21
replace `Box::new` by `into()` when possible

tcoratger:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -762,17 +762,13 @@ fn parse_result(
             };
 
             let value = match ptr {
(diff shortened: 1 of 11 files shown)


edg-l:
I think in this case the `.into()` removes readability of the code so i'm not so inclined to merge this.

tcoratger:
> I think in this case the `.into()` removes readability of the code so i'm not so inclined to merge this.

Feel free to close if you find it not appropriate. As we use `.into()` statements in many other places of the code I think this is more consistent but as you prefer :)

PR #619, opened 2024-05-22
Fix bounded int casts

### LINKED ISSUE #473: Bug: cairo-test integer_tests fail (mostly signed ints)
```cairo
// ./cairo-tests/src/integer_test.cairo

src/integer_test.cairo
1091:// TODO: fails
1296:// TODO: fails
1401:// TODO: fails
1506:// TODO: fails
1611:// TODO: fails
1724:// TODO: fails
1930:    // TODO: fails
```

JulianGCalderon:
closes #473

## Range Bug

Consider the following example: where P is the Felt prime number
```cairo
fn main() -> OneMinusPOnly {
    let felt: felt252 = 1;
    let one_minus_p: BoundedInt<1-P, 1-P> = downcast(felt).expect('felt to bound');
    return one_minus_p;
}
```
The Felt should be castable to the BoundedInt, as 1  = 1-P mod P.

The compiler did not take congruence into account and only checked that 1 was contained in [1-P, 1-P]. As this is false, an error was returned. The solution is to first convert the range to mod P and then check the inclusion. This will return OK as 1 is efectivelly contained in [1, 1].

Note: Euclidian modulo must be used, as a Felt contains only positive integers.

## Bounded Int Representation

In the cairo compiler and the VM, a BoundedInt is treated as a Felt, so representing the BoundedInt as a signed integer complicated the casting operations.

This PR changes the internal representation of BoundedInt to be the same as the Felt.


edg-l (approved):

edg-l (approved):

src/values.rs
```diff
@@ -1339,3 +1375,83 @@ mod test {
...
+                        }
+                        "upper" => {
+                            upper = Some(value);
+                        }
+                        _ => todo!(),
+                    }
+                }
+
+                Ok(Range {
+                    lower: lower.unwrap(),
```
juanbono, line 1449: please handle errors here

src/values.rs
```diff
@@ -1339,3 +1375,83 @@ mod test {
...
+
+                while let Some((field, value)) = map.next_entry()? {
+                    match field {
+                        "lower" => {
+                            lower = Some(value);
+                        }
+                        "upper" => {
+                            upper = Some(value);
+                        }
+                        _ => todo!(),
```
juanbono, line 1444: replace todo!() with an error.

juanbono (changes requested):
Please see comments.

juanbono (approved):

azteca1998 (dismissed):

src/values.rs
```diff
@@ -191,6 +198,28 @@ impl JitValue {
...
+                    range: Range { lower, upper },
+                } => {
+                    let value = value.to_bigint();
+
+                    assert!(lower < upper, "invalid range");
+                    let prime = &PRIME.to_bigint().unwrap();
+                    let lower = lower.rem_euclid(prime);
+                    let upper = upper.rem_euclid(prime);
+                    if lower <= upper {
+                        assert!(lower <= value && value < upper);
```
pefontana, line 212: Can you change the asserts here and replace it with some return Err(..)

pefontana (changes requested):

src/values.rs
JulianGCalderon: done

pefontana (approved):
Nice one!

edg-l (approved):

Issue #634, opened 2024-05-29
Check issue with short_string.cairo program


pefontana:
The short_string.cairo program is running okay in the cairo-vm but has some compilation error when running in Cairo native. Check why and fix it

https://github.com/lambdaclass/cairo_native/blob/89d44a113779307cc48b8577c5f40884995e0a05/tests/tests/cases.rs#L143

```
target/release/cairo-native-run --available-gas 9999999999 --single-file tests/cases/cairo_vm/short_string.cairo
error: Identifier not found.
 --> tests/cases/cairo_vm/short_string.cairo:30:13
        ref your_character
            ^************^

error: ref argument must be a variable.
 --> tests/cases/cairo_vm/short_string.cairo:30:13
        ref your_character
            ^************^

error: Identifier not found.
 --> tests/cases/cairo_vm/short_string.cairo:34:13
        ref your_character
            ^************^

error: ref argument must be a variable.
 --> tests/cases/cairo_vm/short_string.cairo:34:13
        ref your_character
            ^************^

Error: failed to compile: tests/cases/cairo_vm/short_string.cairo
```

Gerson2102:
Can i work on this?

Gerson2102:
I think I'm done, please check sir @pefontana

PR #635, opened 2024-05-30
declaring your_character variable

Gerson2102:
I just added the variable in the cairo program.

## Checklist
- [#634 ] Linked to Github Issue
- [ ] Unit tests added


edg-l:
you should also uncomment the test case so its actually checked now

juanbono (changes requested):
Check that the test is not passing.

Gerson2102:
> Check that the test is not passing.

I dont get it. I just declared a variable, but i did not run anything because that was the problem. That the variable was not declared. And i saw that the CI was failing...

Gerson2102:
And i have the project like this:

image

Is like, is not recognizing my cairo language, and i dont know why.

PR #683, opened 2024-06-10
Basic Stress Test

### LINKED ISSUE #682: Cairo Native Stress Binary
related to #282 

Implement `cairo-native-stress` binary, which loops `n` times over the following sequence:
1. Generate an unique contract. Preferably in memory.
1. Test the contract with AOT, this implies:
    1. Compile it
    1. Cache it
    1. Run it
  
The output of this binary should answer the following questions:
- how many contracts until a segmentation fault?
- how much memory a cached program occupies?

Later on, we should test optimized versions of the cache, like:
- cache with eviction
- cache with write to disk

JulianGCalderon:
related to #682

Adds `cairo-native-stress` that generates, compiles, caches, and runs contracts in order to stress test cairo native.
- Receives by argument the amount rounds to execute.
- In each round it generates, compiles, caches, and runs a new contract
- The compiled dynamic shared libraries are stored on disk
- The cache keeps in memory every executor for every compiled contract
- After each round, it prints a report

## Program Generation

At the start of the program, a dummy program is generated with a hardcoded value (835 in this case) in the return value:
```cairo
#[starknet::contract]
mod Contract {
    #[storage]
    struct Storage {}

    #[external(v0)]
    fn main(self: @ContractState) -> felt252 {
        return 835;
    }
}
```

At the start of each round, the program is traversed to find the hardcoded value and is replaced with a new one. This allows to generate a new valid program with a different return value each round

## Examples

With info logging (20 rounds):
image
With debug logging (5 rounds):
image

## Next Steps

- Export the logs to a JSON or a CSV and plot it
- Try different implementations of the Cache

## Checklist
- [x] Linked to Github Issue

### DIFF
--- a/src/bin/cairo-native-stress.rs
+++ b/src/bin/cairo-native-stress.rs
@@ -1,4 +1,7 @@
+use std::fmt::Display;
+use std::fs::create_dir_all;
 use std::hash::Hash;
+use std::path::Path;
 use std::{collections::HashMap, fs, rc::Rc, time::Instant};
 
 use cairo_lang_sierra::ids::FunctionId;
@@ -24,6 +27,8 @@ struct CliArgs {
 }
 
 fn main() {
+    let cli_args = CliArgs::parse();
+
     tracing::subscriber::set_global_default(
         FmtSubscriber::builder()
             .with_env_filter(EnvFilter::from_default_env())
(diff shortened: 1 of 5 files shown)


src/bin/cairo-native-stress.rs
```diff
@@ -0,0 +1,325 @@
...
+const AOT_CACHE_DIR: &str = ".aot-cache";
+
+/// An unique value hardcoded into the initial contract that it's
+/// used as an anchor point to safely modify it.
+/// It can be any value as long as it's unique in the contract.
+const UNIQUE_CONTRACT_VALUE: u32 = 835;
+
+/// A stress tester for Cairo Native
+///
+/// It Sierra programs compiles with Cairo Native, caches, and executes them with AOT runner.
```
fmoletta, line 44:
```suggestion
/// It compiles Sierra programs with Cairo Native, caches, and executes them with AOT runner.
```

JulianGCalderon: Oh thanks, I totally missed that.

Oppen (approved):

edg-l (approved):

PR #686, opened 2024-06-13
Stress Test: JSON logs

JulianGCalderon:
Allows to save logs to a .jsonl file and plot relevant information from it

- Adds `output` flag for writing logs to a file
- Adds python script that plots the information from the output file
- Adds Makefile commands for running the stress test
- Adds documentation of the stress test in the README

I ran the test in the benchmarking server and got the following information:

server_stress_test

## Checklist
- [ ] Linked to Github Issue


Oppen:
Beautiful plots! 🎉

Oppen (approved):

Cargo.toml
```diff
@@ -36,7 +36,7 @@ required-features = ["scarb"]
 default = ["build-cli", "with-runtime"]
 build-cli = [
   "dep:clap",
-  "dep:tracing-subscriber",
+
```
azteca1998, line 39: The dependency of `tracing-subscriber` should only be present for binaries, not the library. Configuring tracing is the responsibility of the binary, not the library.

JulianGCalderon: Is there a way to specify that a dependency is only used in a particular binary?

JulianGCalderon: Aah, with this: https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies, right? So i should just re add dep:tracing-subscriber to the build-cli feature

JulianGCalderon: I don't remember changing that, maybe it was changed automatically by `cargo add`? Anyway, I already pushed the fix. Thanks

Cargo.toml
```diff
@@ -90,9 +94,7 @@ cairo-lang-starknet-classes = "2.6.3"
 cairo-native-runtime = { version = "0.2.0", path = "runtime", optional = true }
 clap = { version = "4.5", features = ["derive"], optional = true }
 libloading = "0.8.3"
-tracing-subscriber = { version = "0.3", features = [
-  "env-filter",
-], optional = true }
+tracing-subscriber = { version = "0.3.18", features = ["env-filter", "json", "registry"] , optional = true}
```
azteca1998, line 97: Now there's a space missing after `true`.

azteca1998 (dismissed):

Cargo.toml
JulianGCalderon: Fixed

azteca1998 (approved):

edg-l (approved):

PR #690, opened 2024-06-18
Consolidate syscall handler implementations into one (initial pr)

### LINKED ISSUE #649: Consolidate syscall handler trait implementations
There are multiple implementations of the SyscallHandler trait used for various tests. 
It would be good to consolidate them into a single, fully functional, public implementation. 
If any of the implementations is substantially or justifiably different, it can be left out.

- https://github.com/lambdaclass/cairo_native/blob/main/src/utils.rs#L1193
- https://github.com/lambdaclass/cairo_native/blob/main/src/bin/utils/test.rs#L274
- https://github.com/lambdaclass/cairo_native/blob/main/tests/tests/starknet/syscalls.rs#L15
- https://github.com/lambdaclass/cairo_native/blob/main/tests/tests/starknet/keccak.rs#L13
- https://github.com/lambdaclass/cairo_native/blob/main/tests/tests/starknet/secp256.rs#L14
- The DummySyscallHandler in [`starnet.rs`](https://github.com/lambdaclass/cairo_native/blob/main/src/starknet.rs#L299) might possibly be an exemption
- The implementations in the examples may also be an exception, as they hardcode values and leave stubs to keep the examples short. If in the process of consolidating syscall handler implementations a sufficiently functional one is achieved, they could also be replaced in the examples. 

Good documentation of the implementation would be a great contribution.

edg-l:
Partially fi_xes #649 

edit: more syscall handlers can be replaced but it will grow the pr so ill make a followup one

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


edg-l:
more syscall handlers can be replaced but it will grow the pr so ill make a followup one

azteca1998 (approved):

Issue #632, opened 2024-05-29
Improve error handling


pefontana:
Improve the error handling in the repo
Right now we have just one general error type, This is OK, 
https://vscode.dev/github/lambdaclass/cairo_native/blob/main/src/error.rs#L12-L13

But the enum variants inside use mostly strings or aren't so descriptive. We should create a new module for those error variants and use enums instead of strings to describe the error.

check errors
```
Error(String),
ParseAttributeError,
```

adrianvrj:
Hello, is this open for contribution?

I would like to work on this!

adrianvrj:
Hey @pefontana I would love to work on this, could deliver it in 2-3 days!

danielcdz:
Hey @pefontana I would like to help with this one, in the past I made contributions to Nethermind, this can take me about 1-2 days, as I already have some experience with Cairo

Utilitycoder:
```rust
use std::{alloc::LayoutError, num::TryFromIntError};

#[derive(Error, Debug)]
pub enum Error {
    #[error(transparent)]
    LayoutError(#[from] crate::utils::LayoutError),

    #[error(transparent)]
    MlirError(#[from] melior::Error),

    #[error("missing parameter of type '{0}'")]
    MissingParameter(ParameterType),
}

pub enum ParameterType {
    TypeA,
    TypeB,
}
```

Is the above approach in line with the changes you have in mind for the error module? I agree that replacing string-based errors with structured and descriptive enum variants will enhance clarity.

od-hunter:
Hi @pefontana ,Can I be assigned this issue? I’m ready to work.

od-hunter:
Can i hop on this please?

pefontana:
hi @od-hunter !
Yes sure!
i will assign it to you

od-hunter:
@pefontana , Sorry I have been assigned other issues already and I’m working on them, in other not to disappoint you, kindly assign this to someone else, thank you 🙏🏾

gadyrcdz:
Hi @pefontana, if this issue is still available I'd like to help!

adrianvrj:
Hi, is this still open for contribution?
If so, can I hop on this?

hart-venus:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/cairo-native).

### My background and how it can be leveraged
My background is in web development and low level systems, it can be leveraged to create effective and efficient error handling

### How I plan on tackling this issue
[...]

PR #693, opened 2024-06-18
add `CompilerError` and fix `JitValue::BoundedInt` range check

tcoratger:
Part of https://github.com/lambdaclass/cairo_native/issues/632

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


edg-l (dismissed):
thanks!

edg-l (dismissed):

src/error.rs
```diff
@@ -79,6 +80,15 @@ pub enum SierraAssertError {
     Range { ranges: Box<(Range, Range)> },
 }
 
+#[derive(Error, Debug)]
+pub enum CompilerError {
+    #[error("BoundedInt value is out of range: {:?} not within [{:?}, {:?}[", value, range.0, range.1)]
```
Oppen, line 85:
```suggestion
    #[error("BoundedInt value is out of range: {:?} not within [{:?}, {:?}]", value, range.0, range.1)]
```

src/error.rs
```diff
@@ -79,6 +80,15 @@ pub enum SierraAssertError {
...
+    BoundedIntOutOfRange {
+        value: Box<BigInt>,
+        range: Box<(BigInt, BigInt)>,
```
Oppen, line 88: It may be better to `Box` the three values together as one tuple.

Oppen (commented):
There is a little typo in the error message and a little suggestion about the error variant.
It may be a good idea to group the repeated code in added tests into a little setup function, as they look mostly the same. WDYT?

src/error.rs
tcoratger: I did that because the upper range is exclusive and I wanted to highlight that on the error so that the user can understand if for example the problem is that the value is exactly equal to the upper range. It was the mathematical symbol of a range with the upper bound excluded, "[", do you still want to modify it to put a closing bracket "]"?

Oppen: I saw it with parentheses on college for open ranges. Maybe there's more than one notation. I thought it was a typo.

Oppen: (Note that I in fact used a closing parentheses, for the same reason you state, in the suggestion, rather than closing square bracket)

src/error.rs
Oppen: I worry the same, but at the same time there's some risk in having multiple fields in nested errors, as they tend to add up and end up being bigger than the `Ok` variant, which ends up adding overhead to the happy path.
Another option is adding a little struct, but it may be overkill. Leaving it as it is is also a valid option.

tcoratger: Let me know what you prefer and I'll adapt

Oppen: Let's leave it as is.

Oppen (approved):

tcoratger:
@Oppen Anything missing here?

edg-l (approved):

PR #700, opened 2024-06-25
Cleanup values.rs trying to make the code clearer

edg-l:
Some NonNull usage was more a nuisance than useful, `NonNull` is mostly used as a field of a struct to maintain covariance, and here it doesn't really bring anything useful other than adding more lines of code.

There was also a arena alloc with 1 field less on the arrays, i tested if that fixed some stuff from the integrations currently ongoing but it didn't, nevetherless it's a fix.

The cleanup is mainly removing dirty casts and pointers to int to proper methods which maintain pointer provenance, for example instead of casting a ptr of T to a usize to add an offset, we simply use the ptr.bytes_add method, which gives us a pointer offseted by the given bytes.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


azteca1998 (approved):

src/values.rs
```diff
@@ -234,56 +234,44 @@ impl JitValue {
...
-                        );
+                        let target = arena
+                            .alloc_layout(
+                                Layout::new::<*mut ()>() // ptr
+                                    .extend(Layout::new::<u32>())? // start
+                                    .0
+                                    .extend(Layout::new::<u32>())? // end
+                                    .0
+                                    .extend(Layout::new::<u32>())? // capacity
+                                    .0
```
Oppen, line 258: This changes the layout from the original code (adds one field). Are you sure that both are correct?

src/values.rs
```diff
@@ -411,16 +388,20 @@ impl JitValue {
...
-                                value_malloc_ptr.cast().as_ptr(),
+                                value_malloc_ptr.cast(),
                                 elem_layout.size(),
                             );
 
-                            value_map.insert(key, value_malloc_ptr);
+                            value_map.insert(
+                                key,
+                                NonNull::new(value_malloc_ptr)
+                                    .expect("malloc shouldn't return null ptrs")
```
Oppen, line 402: This makes it sound like `malloc` is violating the contract when it's specified to return `NULL` on allocation failure. I'd rephrase the message as this: `"allocation failure"`. It is OK to `panic` since we tolerate that for Rust in general.

src/values.rs
edg-l: yes, the previous was wrong as stated in the pr comment

azteca1998: I can confirm that it's correct as it is now (the array now has 4 fields instead of 3).

Oppen (approved):

pefontana (approved):

PR #707, opened 2024-06-26
Experiment: use libffi instead of custom asm trampoline

edg-l:
work in progress

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -400,7 +400,7 @@ fn invoke_dynamic(
     #[cfg(target_arch = "x86_64")]
     type RetTy = [u64; 2];
     #[cfg(target_arch = "aarch64")]
-    type RetTy = [u64; 2];
+    type RetTy = [u64; 4];
 
     let ret_registers = unsafe {
         closure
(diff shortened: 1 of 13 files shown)


src/executor.rs
```diff
@@ -252,7 +397,18 @@ fn invoke_dynamic(
     let mut ret_registers = [0; 2];
     #[cfg(target_arch = "aarch64")]
     let mut ret_registers = [0; 4];
+    #[cfg(target_arch = "x86_64")]
+    type RetTy = [u64; 2];
+    #[cfg(target_arch = "aarch64")]
+    type RetTy = [u64; 2];
```
Oppen, line 403: Shouldn't this be `4` words long?

edg-l: true, btw this is still a big WIP

src/executor.rs
```diff
@@ -133,6 +134,107 @@ impl<'m> From<JitNativeExecutor<'m>> for NativeExecutor<'m> {
...
+            });
+            let mut data = Vec::new();
+            let mut size = layout.size() * 8;
+            dbg!(size);
+            while size > 0 {
+                if size >= 64 {
+                    data.push(Type::i64());
+                    size -= 64;
+                } else if size >= 32 {
+                    data.push(Type::i64());
```
tcoratger, line 199: Maybe i32 instead of i64 here no?

PR #720, opened 2024-07-05
Fix: cairo_native swaps the high and low fields of U256 compared to Cairo's u256

### LINKED ISSUE #715: Bug: cairo_native swaps the high and low fields of `U256` compared to Cairo's `u256`
The `hi` and `lo` field of `U256` within cairo native doesn't correspond to `high` and `low` fields of `u256` in cairo; they are swapped. 

# Reproducer

In the starknet example `keccak` is implemented as:
```rust
    fn keccak(
        &mut self,
        input: &[u64],
        _gas: &mut u128,
    ) -> SyscallResult<cairo_native::starknet::U256> {
        println!("Called `keccak({input:?})` from MLIR.");
        Ok(U256 {
            hi: 0,
            lo: 1234567890,
        })
    }
```

Invoking this from cairo as follows:
```cairo
    #[external(v0)]
    fn echo(ref self: ContractState, value: felt252) -> felt252 {
        let input : Array::<u64> = Default::default();
        let output = starknet::syscalls::keccak_syscall(input.span()).unwrap();
        if output.low == 0 && output.high == 1234567890 {
            panic_with_felt252('arguments swapped');
        }
        output.low.into()
    }
```
leads to the panic 'arguments swapped'. 

The reproducer can be cloned from [here](https://github.com/NethermindEth/cairo_native/tree/u256-ordening-bug) and executed with `cargo r --example starknet`.

edg-l:
Fixes #715

I introduced this when i added the corelib tests, instead of changing the field order of the u256, instead i swapped the tests and starknet syscall impl, but that was wrong, now its fixed

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/starknet_stub.rs
```diff
@@ -258,8 +258,8 @@ impl StarknetSyscallHandler for &mut StubSyscallHandler {
         // state[0] and state[1] conform the hash_high (u128)
         // state[2] and state[3] conform the hash_low (u128)
         SyscallResult::Ok(U256 {
-            lo: state[2] as u128 | ((state[3] as u128) << 64),
-            hi: state[0] as u128 | ((state[1] as u128) << 64),
+            hi: state[2] as u128 | ((state[3] as u128) << 64),
+            lo: state[0] as u128 | ((state[1] as u128) << 64),
```
azteca1998, line 262: Since we're ordering them lo then hi in the definition maybe we should keep the same order here? It also makes sense since we'll be accessing `state` in order.

azteca1998 (approved):

src/starknet_stub.rs
pefontana: I agree with azteca1998

juanbono (dismissed):

pefontana (dismissed):

pefontana (approved):

juanbono (approved):

PR #724, opened 2024-07-08
Fix dict snapshot

### LINKED ISSUE #723: Returning snapshot of dict causes double free
```
fn main() -> @Felt252Dict<u32> {
    let mut dict: Felt252Dict<u32> = Default::default();
                dict.insert(2, 1_u32);

    @dict
}
```

```
free(): double free detected in tcache 2
Aborted
```

I think we should also handle `snapshot_take` for dicts

edg-l:
Also frees the allocated dict values, beforehand they werent and probs leaked.

Fixes #723 

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/types/felt252_dict.rs
+++ b/src/types/felt252_dict.rs
@@ -6,13 +6,18 @@
 //! used to count accesses to the dictionary. The type is interacted through the runtime functions to
 //! insert, get elements and increment the access counter.
 
+use std::cell::Cell;
+
 use super::WithSelf;
 use crate::{
+    block_ext::BlockExt,
     error::Result,
     libfuncs::LibfuncHelper,
     metadata::{
-        runtime_bindings::RuntimeBindingsMeta, snapshot_clones::SnapshotClonesMeta, MetadataStorage,
+        realloc_bindings::ReallocBindingsMeta, runtime_bindings::RuntimeBindingsMeta,
+        snapshot_clones::SnapshotClonesMeta, MetadataStorage,
     },
+    types::TypeBuilder,
 };
 use cairo_lang_sierra::{
     extensions::{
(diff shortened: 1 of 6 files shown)


src/types/felt252_dict.rs
```diff
@@ -28,22 +34,79 @@ pub fn build<'ctx>(
...
+    _registry: &ProgramRegistry<CoreType, CoreLibfunc>,
+    entry: &'this Block<'ctx>,
+    location: Location<'ctx>,
+    helper: &LibfuncHelper<'ctx, 'this>,
+    metadata: &mut MetadataStorage,
+    _info: WithSelf<InfoAndTypeConcreteType>,
+    src_value: Value<'ctx, 'this>,
+) -> Result<(&'this Block<'ctx>, Value<'ctx, 'this>)> {
+    // This snapshot take assumes the value types in a felt are always copy and don't need deep cloning.
+    // So far (atleast up to 2.6.4) you cannot make a felt dict of array values, so this holds.
```
juanbono, line 66: and what about structs?

juanbono: it can hold arbitrary types via a Nullable<T> value. So if we have a struct `A` (or an array, another dict, etc), `FeltDict252<Nullable<A>>` is valid.

juanbono (changes requested):

edg-l:
Converted to draft as its missing this use case

```
fn main() -> @Felt252Dict<Nullable<Array<u32>>> {
    let mut dict: Felt252Dict<Nullable<Array<u32>>> = Default::default();
    dict.insert(2, NullableTrait::new(array![0]));

    @dict
}
```

src/types/felt252_dict.rs
Oppen: This comment contradicts the findings in #723. Indeed, you can create and take snapshots from a `DictFelt<Array<Felt>>`.

juanbono (approved):

azteca1998 (approved):

PR #731, opened 2024-07-16
Refactor `ArgumentMapper` to fix bugs in ARM.

azteca1998:
- Fixes `to_jit()` for arrays (we didn't notice because we weren't using it before).
  - Completely replaces `ArgumentMapper` with a more intuitive system that can integrate multiple architectures easily.
  - Fixes the `Failed to deserialize param #1` error.
  - Fixes a bug in our debug utils `print_felt252`.
  - Makes `JitValue::Felt(_)`'s `std::fmt::Debug` implementation print the real value instead of its Montgomery form limbs.
  - Adds a test to check if the circumstances that caused this bug happens in the future.

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -1,0 +1,177 @@
+use crate::{
+    starknet::{ArrayAbi, U256},
+    types::TypeBuilder,
+    values::JitValue,
+};
+use bumpalo::Bump;
+use cairo_lang_sierra::{
+    extensions::{
+        core::{CoreLibfunc, CoreType, CoreTypeConcrete},
+        starknet::{secp256::Secp256PointTypeConcrete, StarkNetTypeConcrete},
+    },
+    ids::ConcreteTypeId,
+    program_registry::ProgramRegistry,
+};
+use std::ptr::NonNull;
+
+mod aarch64;
+mod x86_64;
+
+/// Implemented by all supported argument types.
+pub trait AbiArgument {
+    fn to_bytes(&self, buffer: &mut Vec<u8>);
+}
+
+pub struct JitValueWithInfoWrapper<'a> {
+    pub value: &'a JitValue,
+    pub type_id: &'a ConcreteTypeId,
+    pub info: &'a CoreTypeConcrete,
+
+    pub arena: &'a Bump,
+    pub registry: &'a ProgramRegistry<CoreType, CoreLibfunc>,
+}
+
+impl<'a> JitValueWithInfoWrapper<'a> {
+    fn map<'b>(
+        &'b self,
+        value: &'b JitValue,
+        type_id: &'b ConcreteTypeId,
+    ) -> JitValueWithInfoWrapper<'b>
+    where
+        'b: 'a,
+    {
+        Self {
+            value,
+            type_id,
+            info: self.registry.get_type(type_id).unwrap(),
+            arena: self.arena,
+            registry: self.registry,
+        }
+    }
+}
+
+impl<'a> AbiArgument for JitValueWithInfoWrapper<'a> {
+    fn to_bytes(&self, buffer: &mut Vec<u8>) {
+        match (self.value, self.info) {
+            (
+                value,
+                CoreTypeConcrete::Box(info)
+                | CoreTypeConcrete::NonZero(info)
+                | CoreTypeConcrete::Nullable(info)
+                | CoreTypeConcrete::Snapshot(info),
+            ) => {
+                // TODO: Allocate and use to_jit().
+                self.map(value, &info.ty).to_bytes(buffer)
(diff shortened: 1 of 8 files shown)


edg-l (approved):

src/arch.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub value: &'a JitValue,
+    pub type_id: &'a ConcreteTypeId,
+    pub info: &'a CoreTypeConcrete,
+
+    pub arena: &'a Bump,
+    pub registry: &'a ProgramRegistry<CoreType, CoreLibfunc>,
+}
+
+impl<'a> JitValueWithInfoWrapper<'a> {
+    fn map<'b>(
```
juanbono, line 35: can you add some docs to this function? It's short but I think it can help.

src/arch/aarch64.rs
```diff
@@ -0,0 +1,229 @@
...
+        } else {
+            align_to(buffer, get_integer_layout(128).align());
+            buffer.extend_from_slice(&self.to_ne_bytes());
+        }
+    }
+}
+
+impl AbiArgument for Felt {
+    fn to_bytes(&self, buffer: &mut Vec<u8>) {
+        if buffer.len() < 56 {
```
juanbono, line 124: can you add a small comment here why it checks against 56? I think it can improve readability

src/arch/aarch64.rs
```diff
@@ -0,0 +1,229 @@
...
+        } else {
+            align_to(buffer, get_integer_layout(252).align());
+            buffer.extend_from_slice(&self.to_bytes_le());
+        }
+    }
+}
+
+impl AbiArgument for U256 {
+    fn to_bytes(&self, buffer: &mut Vec<u8>) {
+        if buffer.len() < 56 {
```
juanbono, line 135: same here

mpaulucci:
🐐

src/arch/aarch64.rs
```diff
@@ -0,0 +1,229 @@
...
+            align_to(buffer, get_integer_layout(256).align());
+            buffer.extend_from_slice(&self.lo.to_le_bytes());
+            buffer.extend_from_slice(&self.hi.to_le_bytes());
+        }
+    }
+}
+
+impl AbiArgument for [u8; 31] {
+    fn to_bytes(&self, buffer: &mut Vec<u8>) {
+        if buffer.len() < 56 {
```
juanbono, line 148: Please add a comment here with the reason why here it checks with 56

src/arch/x86_64.rs
```diff
@@ -0,0 +1,229 @@
...
+use num_traits::ToBytes;
+use starknet_types_core::felt::Felt;
+
+fn align_to(buffer: &mut Vec<u8>, align: usize) {
+    buffer.resize(buffer.len().next_multiple_of(align), 0);
+}
+
+impl AbiArgument for u8 {
+    fn to_bytes(&self, buffer: &mut Vec<u8>) {
+        if buffer.len() < 48 {
```
juanbono, line 14: same here

src/executor.rs
```diff
@@ -219,34 +216,42 @@ fn invoke_dynamic(
...
+
+                arena: &arena,
+                registry,
+            }
+            .to_bytes(&mut invoke_data),
         }
     }
 
+    // Pad invoke data to the 16 byte boundary avoid segfaults.
+    // TODO: Test if the resize must be inserted before or after the data.
```
juanbono, line 242: we should create issues to track these TODOs

juanbono (changes requested):
Amazing work!!!!!! 🚀 , I left some comments, most of them are about adding some docs on the AbiArgument implementations.

src/executor.rs
azteca1998: Some (all?) of them are no longer applicable.

src/arch.rs
azteca1998: Done. Not to the impl but in the trait.

src/arch/aarch64.rs
azteca1998: Done. I added at the module level because the `64` and `56` numbers are used in various places.

src/arch/aarch64.rs
azteca1998: Done. I also mentioned why `bytes31` also uses the same splitting rules as other integers.

src/arch/x86_64.rs
azteca1998: Done.

edg-l (approved):

juanbono (approved):

PR #733, opened 2024-07-17
Fix `Box<T>` and `Nullable<T>` as entry point arguments.

azteca1998:
Something I missed from the previous fix.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -57,20 +57,46 @@ impl<'a> JitValueWithInfoWrapper<'a> {
 impl<'a> AbiArgument for JitValueWithInfoWrapper<'a> {
     fn to_bytes(&self, buffer: &mut Vec<u8>) {
         match (self.value, self.info) {
-            (value, CoreTypeConcrete::Box(_)) => {
+            (value, CoreTypeConcrete::Box(info)) => {
                 let ptr = value
                     .to_jit(self.arena, self.registry, self.type_id)
                     .unwrap();
-                ptr.as_ptr().to_bytes(buffer);
+
+                let layout = self
+                    .registry
+                    .get_type(&info.ty)
+                    .unwrap()
+                    .layout(self.registry)
+                    .unwrap();
+                let heap_ptr = unsafe {
+                    let heap_ptr = libc::malloc(layout.size());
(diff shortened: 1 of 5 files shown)


edg-l (approved):

src/arch.rs
```diff
@@ -57,14 +57,49 @@ impl<'a> JitValueWithInfoWrapper<'a> {
...
-                CoreTypeConcrete::Box(info)
-                | CoreTypeConcrete::NonZero(info)
-                | CoreTypeConcrete::Nullable(info)
-                | CoreTypeConcrete::Snapshot(info),
-            ) => {
-                // TODO: Allocate and use to_jit().
+            (value, CoreTypeConcrete::Box(info)) => {
+                let ptr = value
+                    .to_jit(self.arena, self.registry, self.type_id)
+                    .unwrap();
```
pefontana, line 63: Are these unwraps safe?
Should be implement an error handling here? Maybe we can make it in another PR if it is necesarry

src/arch.rs
```diff
@@ -57,14 +57,49 @@ impl<'a> JitValueWithInfoWrapper<'a> {
...
+
+                heap_ptr.to_bytes(buffer);
+            }
+            (value, CoreTypeConcrete::Nullable(info)) => {
+                if matches!(value, JitValue::Null) {
+                    null::<()>().to_bytes(buffer);
+                } else {
+                    let ptr = value
+                        .to_jit(self.arena, self.registry, self.type_id)
+                        .unwrap();
```
pefontana, line 85: same here

pefontana (changes requested):

src/arch.rs
azteca1998: Ideally, yes. It could technically fail, but it'll most likely be because of an invalid Sierra program or internal Cairo error (the libraries we're using).
We can turn it into an error later on along with all the others once we fix our current bugs.

src/arch.rs
pefontana: #748

pefontana (approved):

PR #750, opened 2024-08-02
Update version of `starknet-types-core` and remove unused dependencies  

varex83:
While integrating cairo `2.7.0-rc.3` to the [blockifier](https://github.com/NethermindEth/blockifier/pull/121), we've got a dependency error with incorrect `starknet-types-core` version. Also while executing new CI after the update, we've found out that `cairo-lang-runner` is not being used in the `cairo-native-runtime` crate, so we've removed it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -989,7 +989,6 @@ dependencies = [
 name = "cairo-native-runtime"
 version = "0.2.0"
 dependencies = [
- "cairo-lang-runner",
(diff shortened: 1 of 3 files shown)


edg-l:
I think you need to re-do or rebase the pr, since we updated the target branch with the release 2.7.0

varex83:
@edg-l can you re-check it please?

edg-l (approved):

PR #768, opened 2024-08-26
make u128 from felt impl match cairo to casm one

edg-l:
https://github.com/starkware-libs/cairo/blob/b47c0fafae61a67e8944943fd2f6711c2db2e56d/crates/cairo-lang-sierra-to-casm/src/invocations/int/unsigned128.rs#L234

maybe our current impl is buggy and may help with the overflow bugs

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/libfuncs/uint128.rs
```diff
@@ -194,29 +195,33 @@ pub fn build_from_felt252<'ctx, 'this>(
...
+        value,
         IntegerType::new(context, 128).into(),
         location,
     ))?;
 
-    let msb_bits = entry.append_op_result(arith::shrui(arg1, k128, location))?;
-    let msb_bits = entry.append_op_result(arith::trunci(
-        msb_bits,
+    let value_if_of = entry.append_op_result(arith::divui(value, k_u128_max_1, location))?;
+    let overflow = entry.append_op_result(arith::remui(value, k_u128_max_1, location))?;
```
azteca1998, line 216: Isn't this the same we had earlier, but less efficient? I think the only thing that really changed is the `is_wide` comparison, which is now properly implemented (we're comparing against `u128::MAX + 1` instead of `128`).

Could you try if just shifting and truncating works now too?

azteca1998 (approved):

azteca1998 (approved):

pefontana (approved):

PR #772, opened 2024-08-27
Improve error handling

### LINKED ISSUE #632: Improve error handling

RuneRogue:
Updated:
✅src/error.rs

All error types are now within a dedicated errors module. Instead of using strings, error details are captured by enums like MissingParameterError and UnexpectedValueError. The error messages are now more structured and informative, making it easier to understand the error context.

Fixed #632


edg-l:
We believe our error handling is good right now, that issue was old and a bit outdated, sorry for the inconvenience.

PR #782, opened 2024-09-05
Contract Executor

### LINKED ISSUE #781: Add contract-specific Executor with improved API
The current AOTExecutor carries some information with it that is not strictly necessary for contract execution, namely the program registry and the metadata. 

It would be beneficial to have a more streamlined executor with baked-in assumptions allowed by only taking into account starknet contracts to be executed. This `ContractExecutor` executor would not need the program registry nor the metadata. 

The API for this `ContractExecutor` should have 
- a `new` function to instantiate it from a `NativeModule`
- a `load` function to instantiate it from an already compiled `NativeModule`
- a `run` function to run a function on an alread instantiated executor (parameters should be the function name should be passed in, and the function id obtained from the mapping, the args, and gas)

The gas metadata is used to check whether the program to be executed requires the gas builtin, which is guaranteed in the case of all starknet contracts, and the available gas is always required to be passed. In addition, all starknet contracts use the same fixed set of builtins. 

In addition, information that is still needed should be available to the executor on instantiation, and should be stored between executor instance creations in a format that is amenable to serialization, so that the program registry or other information from the sierra representation of the program is not required on loading. 
Right now this is mostly the mapping of entry points to function ids.

Thus, the output of compilation in AOT mode (perhaps as part of the `module_to_object` functionality)  should include a json object on disk with this mapping, that can be quickly (de)serialized on loading the dynamic library, so that the call to `invoke_dynamic` does not require further processing.

edg-l:
Fixes #781 

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/executor/ce.rs
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    /// Clone this executor, can't be the trait because it can fail.
+    ///
+    /// It loads from the stored path the library and json file to "clone" the executor.
+    pub fn clone_executor(&self) -> Result<Self, Error> {
+        let mut x = Self::load(&self.path)?;
+        x.entry_points_info = self.entry_points_info.clone();
+        Ok(x)
+    }
```
azteca1998, line 176: Why not `Arc<_>` it? Then it would be cloneable.

azteca1998 (dismissed):

azteca1998 (approved):

src/libfuncs/circuit.rs
```diff
@@ -1382,12 +1382,12 @@ mod test {
...
 
-        dbg!(Felt::from_raw([
+        Felt::from_raw([
             576460752303419696,
             18446744073709551615,
             18446744073709551615,
-            18446744073709551393
-        ]));
+            18446744073709551393,
+        ]);
```
JulianGCalderon, line 1390: Oh, I think is accidentally left this debug statement when I implemented circuits. It should be safe to remove (the whole statement).

pefontana (approved):

PR #783, opened 2024-09-09
Add trace dump contracts 2

JulianGCalderon:


runtime/Cargo.toml
```diff
@@ -26,6 +26,8 @@ starknet-crypto = "0.7.1"
 starknet-curve = "0.5.0"
 lazy_static = "1.5.0"
 rand = "0.8.5"
+num-bigint = "0.4.4"
+num-traits = "0.2"
```
azteca1998, line 30: Are those deps only used in the trace dump? If so they should be behind the trace dump feature.

JulianGCalderon: Done

PR #787, opened 2024-09-10
Fix second tail recursive transformation bug.

azteca1998:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/compiler.rs
```diff
@@ -1283,3 +1310,9 @@ fn generate_entry_point_wrapper<'c>(
     ));
     Ok(())
 }
+
+#[derive(Clone, Debug)]
+enum StatementCompileResult<T> {
+    Processed(T),
+    Deferred,
+}
```
JulianGCalderon, line 1318: I think this is a bit tricky to understand (it took me a while). Could you add some documentation on how the `StatementCompileResult` integrates with `foreach_statement_in_function`?

JulianGCalderon (approved):
Nice one @azteca1998!! I've tested this with starknet-replay and seems to fix all overflow bugs.

src/compiler.rs
azteca1998: Done.

edg-l (approved):

PR #809, opened 2024-09-25
Add issues templates

pefontana:


edg-l (dismissed):

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+What Cairo Native commit are you using
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+* What project are you running
+* Which Cairo, Sierra, or Starknet transaction are you running
+* What Error are you getting
```
azteca1998, line 23: Could we also request the exact commands for reproducing the bug as well as a minimal example in cases where it makes sense? (maybe a contract is impossible to reduce but if it's a libfunc there should be a minimal example).

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. MacOS, Ubuntu 22.04 LTS]
```
azteca1998, line 30: They should also provide their architecture (x86_64 or aarch64), not only their OS since we have quite a bit of architecture-dependent code.

.github/ISSUE_TEMPLATE/bug_report.md
pefontana: Sure!
Good advice @azteca1998

.github/ISSUE_TEMPLATE/bug_report.md
pefontana: Good!
done

edg-l (approved):

PR #815, opened 2024-09-26
Fix clone and drop implementations for enums, structs and snapshots

azteca1998:
**Changes**
  - Rename a few things.
  - Remove `SnapshotClonesMeta`.
  - Remove old drop implementation system.
  - Add `DupOverridesMeta` (successor of `SnapshotClonesMeta`) and `DropOverridesMeta`.
  - Make all types that require it use the new dup and drop systems.
  - Refactor some dict methods to make them work with the new dup and drop system.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -134,73 +134,72 @@ pub unsafe extern "C" fn cairo_native__libfunc__hades_permutation(
 }
 
 /// Felt252 type used in cairo native runtime
-pub type FeltDict = (HashMap<[u8; 32], NonNull<std::ffi::c_void>>, u64);
-
-/// Allocates a new dictionary. Internally a rust hashmap: `HashMap<[u8; 32], NonNull<()>`
-///
-/// # Safety
-///
-/// This function is intended to be called from MLIR, deals with pointers, and is therefore
-/// definitely unsafe to use manually.
-#[no_mangle]
-pub unsafe extern "C" fn cairo_native__alloc_dict() -> *mut std::ffi::c_void {
-    Box::into_raw(Box::<FeltDict>::default()) as _
+#[derive(Debug, Default)]
+pub struct FeltDict {
+    pub inner: HashMap<[u8; 32], NonNull<std::ffi::c_void>>,
+    pub count: u64,
 }
 
-/// Frees the dictionary.
+/// Allocate a new dictionary.
 ///
 /// # Safety
 ///
 /// This function is intended to be called from MLIR, deals with pointers, and is therefore
 /// definitely unsafe to use manually.
 #[no_mangle]
-pub unsafe extern "C" fn cairo_native__dict_free(ptr: *mut FeltDict) {
-    let mut map = Box::from_raw(ptr);
+pub unsafe extern "C" fn cairo_native__dict_new() -> *mut FeltDict {
+    Box::into_raw(Box::<FeltDict>::default())
+}
+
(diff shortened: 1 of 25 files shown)


src/types/struct.rs
```diff
@@ -57,12 +63,130 @@ pub fn build<'ctx>(
...
+
+        // The following unwrap is unreachable because the registration logic will always insert it.
+        let values = metadata
+            .get::<DupOverridesMeta>()
+            .unwrap()
+            .invoke_override(context, &entry, location, member_id, member_val)?;
+
+        src_value = entry.insert_value(context, location, src_value, values.0, idx)?;
+        dst_value = entry.insert_value(context, location, dst_value, values.1, idx)?;
+    }
```
JulianGCalderon, line 158: This will only dup members that have an override, while if a member is copyable, i won't be cloned to the new struct. Am I right, or I'm missing something?

azteca1998: This snippet will:
  - For each member in the struct (regardless of whether it's copyable or requires clone):
    - Build its type (makes sure the clone implementation is registered.
    - Extract the value.
    - Invoke the clone function. If registered, then it'll be cloned. Otherwise the same value is returned twice (therefore copying it).
    - Store the results into the original and cloned structs. The original is also stored just in case we need it in the future.
    - Update `src_value` and `dst_value` variables so that the next member builds from them.

This means that it'll work for both copyable and cloneable members. I'm not sure if this answers your question.

JulianGCalderon: Aaah, I see! Thanks for the explanation

runtime/src/lib.rs
```diff
@@ -134,73 +134,72 @@ pub unsafe extern "C" fn cairo_native__libfunc__hades_permutation(
...
 /// # Safety
 ///
 /// This function is intended to be called from MLIR, deals with pointers, and is therefore
 /// definitely unsafe to use manually.
 #[no_mangle]
-pub unsafe extern "C" fn cairo_native__dict_free(ptr: *mut FeltDict) {
-    let mut map = Box::from_raw(ptr);
+pub unsafe extern "C" fn cairo_native__dict_drop(
+    ptr: *mut FeltDict,
+    drop_fn: Option<extern "C" fn(*mut std::ffi::c_void)>,
```
edg-l, line 163: I would add a comment here, clarifying that this Option is ffi safe due to this: https://doc.rust-lang.org/nomicon/ffi.html#the-nullable-pointer-optimization

edg-l (dismissed):

runtime/src/lib.rs
azteca1998: Done

src/libfuncs/array.rs
```diff
@@ -491,6 +491,9 @@ pub fn build_get<'ctx, 'this>(
         // TODO: Support clone-only types (those that are not copy).
         valid_block.memcpy(context, location, elem_ptr, target_ptr, elem_size);
 
+        // TODO: Drop array values.
```
edg-l, line 494: there is a missing drop for array values

edg-l (approved):

src/libfuncs/array.rs
edg-l: we can do this in another pr

edg-l (approved):

edg-l (approved):

JulianGCalderon (approved):

PR #833, opened 2024-10-04
Add malloc tracing and fix more memory leaks.

azteca1998:
- Fix all memory leaks found JIT-run tests (including the Cairo corelib tests).
- Refactor some libfuncs to remove memory leaks.
- Refactor dictionaries.
  - Removed `dict_insert` runtime function.
  - Delayed element allocation until actually needed.
- Add memory allocation tracing gated behind a feature.
  - Make the program crash if there was a memory leak (again, behind a feature).

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/starknet.rs
```diff
@@ -251,10 +251,10 @@ pub trait StarknetSyscallHandler {
 
     fn sha256_process_block(
         &mut self,
-        prev_state: &[u32; 8],
-        current_block: &[u32; 16],
+        state: &mut [u32; 8],
+        block: &[u32; 16],
         remaining_gas: &mut u128,
-    ) -> SyscallResult<[u32; 8]>;
+    ) -> SyscallResult<()>;
```
enitrat, line 257: q: why was this modified? isn't this going to be a problem for upstream blockifier? (I had to adapt my types so i'm wondering)

azteca1998: There are various reasons:
  - The technical reason is that internally the state is represented as a box. We receive a box and return a box so we're just reusing the allocation.
  - All libraries I've seen mutate the state instead of returning a new one.
  - We can simplify the MLIR code by avoiding (stack) `alloca`s and `memcpy`s.

Before this we `malloc`'d a new state, passed the old state's pointer as the current state and the new pointer as the return value's location, which meant that the previous state's allocation got leaked. Now it doesn't get leaked and we do not need extra allocations.

Our stub syscall handler has an example implementation with the changes to adapt it [here](https://github.com/lambdaclass/cairo_native/pull/833/files/b73e3fa8927eb284d362594aeaff900e28636a8d#diff-b9b3ae981ddf1b61c79d36505c39e96f0ad175b042d384f1e1d9d637a1034991R823-R837).

src/types/felt252_dict_entry.rs
```diff
@@ -36,12 +35,12 @@ pub fn build<'ctx>(
     _metadata: &mut MetadataStorage,
     _info: WithSelf<InfoAndTypeConcreteType>,
 ) -> Result<Type<'ctx>> {
+    // TODO: Custom drop (dup?) for dict entries.
```
edg-l, line 38: is this TODO still relevant? same for the other todo

edg-l (approved):

src/types/felt252_dict_entry.rs
azteca1998: Maybe. The dict entry owns the dict, so it may need to be dropped, but I'm not sure it's possible to drop a dict entry (it's handled internally). Let me try and break it so that it drops an entry and see what happens.

Also, where is the other drop?

runtime/src/lib.rs
```diff
@@ -157,23 +165,26 @@ pub unsafe extern "C" fn cairo_native__dict_new() -> *mut FeltDict {
...
-            drop_fn(entry.as_ptr());
-        }
+    for entry in dict.inner.into_values() {
+        if !entry.is_null() {
+            if let Some(drop_fn) = drop_fn {
+                drop_fn(entry);
+            }
 
-        libc::free(entry.as_ptr());
+            // TODO: Use the mem tracing subsystem (avoid false memory leaks).
```
edg-l, line 185: here

JulianGCalderon (commented):
I've tested this and works OK, although AOT tests fail when using the feature, should we disable those tests when running with `with-mem-tracing` feature?

azteca1998:
> although AOT tests fail when using the feature

Yes, this feature depends on being able to override `malloc`, `realloc` and `free` within the JIT only, not the entire program, which makes it very hard to do in AOT.

> should we disable those tests when running with with-mem-tracing feature?

In the CI? No, it's better to disable the feature rather than disabling some tests since this feature is for debugging and whereas the tests are for production.
Locally? I'd say there's no need and it may cause confusion when the test you're trying to run suddenly doesn't exist anymore when you enable memory tracing.

runtime/src/lib.rs
azteca1998: This one was old and no longer applied.

src/types/felt252_dict_entry.rs
azteca1998: Apparently you can neither drop nor dup a dict entry so those are not necessary.
What you can do, however, is return a dict entry from the entrypoint, but I haven't implemented it because it may never happen (it's internal stuff). It won't affect contracts in any way since they use serializaztion.

edg-l (approved):

pefontana (approved):

PR #837, opened 2024-10-08
Builtin costs rework

### LINKED ISSUE #836: Provide way to set the builtin gas cost.
Blockifier needs to be able to set the gas cost per builtin as is done [here](https://github.com/starkware-libs/sequencer/blob/89243a133938719a7c715238ccd625bf08f2a4f4/crates/blockifier/src/execution/entry_point_execution.rs#L206-L221) for the vm, since these values can change per version.

edg-l:
Fixes #836 

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/compiler.rs
```diff
@@ -135,6 +135,30 @@ pub fn compile(
...
+                context,
+                region,
+                TypeAttribute::new(llvm::r#type::pointer(context, 0)),
+                StringAttribute::new(context, "builtin_costs"),
+                Attribute::parse(context, "#llvm.linkage<external>").unwrap(),
+                location,
+            )
+            .into(),
+        );
+    }
```
azteca1998, line 160: Does it have to be a global? It won't cause an issue with neither the AOT (shared libraries only) or JIT, but it could be problematic when compiling a standalone object from multiple cairo programs due to symbol redefinitions.

src/executor.rs
```diff
@@ -141,6 +144,25 @@ fn invoke_dynamic(
...
+    // https://github.com/starkware-libs/sequencer/blob/1b7252f8a30244d39614d7666aa113b81291808e/crates/blockifier/src/execution/entry_point_execution.rs#L208
+    let builtin_costs_array: &[u64] = &[
+        token_gas_cost(CostTokenType::Const) as u64,
+        token_gas_cost(CostTokenType::Pedersen) as u64,
+        token_gas_cost(CostTokenType::Bitwise) as u64,
+        token_gas_cost(CostTokenType::EcOp) as u64,
+        token_gas_cost(CostTokenType::Poseidon) as u64,
+        token_gas_cost(CostTokenType::AddMod) as u64,
+        token_gas_cost(CostTokenType::MulMod) as u64,
+    ];
```
azteca1998, line 157: Since order matters and every field has a different meaning, maybe we could use a struct instead? Then they'll have names associated with the number.

src/executor.rs
```diff
@@ -166,6 +188,14 @@ fn invoke_dynamic(
...
                     .to_bytes(&mut invoke_data)?;
             }
+            CoreTypeConcrete::BuiltinCosts(_) => {
+                // This builtin should never be an argument but just in case.
+                if let Some(builtin_costs) = builtin_costs {
+                    builtin_costs.to_bytes(&mut invoke_data)?;
+                } else {
+                    (builtin_costs_array.as_ptr()).to_bytes(&mut invoke_data)?;
+                }
+            }
```
azteca1998, line 198: This match branch will make the changes below (line 282 after changes) unreachable.

src/types/builtin_costs.rs
```diff
@@ -1,4 +1,7 @@
 //! # Builtin costs type
+//!
+//! A ptr to a list of u64, this list will not change at runtime in size and thus we only really need to store the pointer,
+//! it can be allocated on the stack on rust side and passed.
```
azteca1998, line 4: Does this pointer point to the global declared earlier? If so, why do we need the global?

programs/benches/factorial_2M.c
```diff
@@ -25,6 +26,10 @@ int main()
 {
     factorial_return_values_t return_values;
 
+    uint64_t BuiltinCosts[7] = {1, 4050, 583, 4085, 491, 230, 604};
```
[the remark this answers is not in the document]
edg-l: they are the current token gas costs, for the benchmark they dont really matter if they are updated / actual but they have to exist

src/compiler.rs
edg-l: Yes it has to be a global, the builtin costs is not passed anywhere as an input.

> but it could be problematic when compiling a standalone object from multiple cairo programs due to symbol redefinitions.

Is this something we do? I think its out of scope anyway, projects / contracts always end up being 1 sierra file and thus 1 library

src/types/builtin_costs.rs
edg-l: No, the global has a pointer, which points to the list.

This type represents the pointer to the list, to get the BuiltinCosts type for the libfunc that retrieves it (and for one of the withdraws that doesnt have a builtincosts argument) you need to have a global to get the ptr to the list

src/executor.rs
edg-l: thats a good idea

src/executor.rs
edg-l: i dont see it, can you explain in more detail?

src/compiler.rs
azteca1998: The ones run with cairo-native-run yes, but we can also compile it into an object file and link it with other programs. I guess it's fine for now since that use case is so rare and not our priority. Also I don't see how we could do this without a global or a big refactor.

src/executor.rs
azteca1998: Nevermind I'm blind.

src/types/builtin_costs.rs
azteca1998: Understood. Thank you.

azteca1998 (approved):

src/executor.rs
```diff
@@ -249,26 +267,38 @@ fn invoke_dynamic(
...
-                        CoreTypeConcrete::Poseidon(_) => builtin_stats.poseidon = value,
-                        CoreTypeConcrete::SegmentArena(_) => builtin_stats.segment_arena = value,
-                        CoreTypeConcrete::RangeCheck96(_) => builtin_stats.range_check_96 = value,
-                        CoreTypeConcrete::Circuit(CircuitTypeConcrete::AddMod(_)) => {
-                            builtin_stats.circuit_add = value
-                        }
-                        CoreTypeConcrete::Circuit(CircuitTypeConcrete::MulMod(_)) => {
-                            builtin_stats.circuit_mul = value
+                    if let CoreTypeConcrete::BuiltinCosts(_) = type_info {
+                        // todo: should we use this value?
```
pefontana, line 271: @edg-l Can you create an issue to track this?

pefontana (approved):

PR #841, opened 2024-10-10
Resolve `CAIRO_NATIVE_RUNTIME_LIBRARY` relative path

rodrigo-pino:
Simple PR to resolve the `CAIRO_NATIVE_RUNTIME_LIBRARY` when it is a relative path so that it can work better with STARKNET approach to global vars (e.g. doing "./native_runtime.a") will turn into "/usr/somewhere/over/the/rainbow/native_runtime.a".

Looking forward to hear your thoughts!

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/scripts/bench-hyperfine.sh
+++ b/scripts/bench-hyperfine.sh
@@ -90,7 +90,7 @@ run_bench() {
         -o "$OUTPUT_DIR/$base_name-march-native" \
         >> /dev/stderr
 
-    CAIRO_NATIVE_RUNTIME_LIBDIR="$ROOT_DIR/target/release" hyperfine \
+     hyperfine \
         --warmup 3 \
         --export-markdown "$OUTPUT_DIR/$base_name.md" \
         --export-json "$OUTPUT_DIR/$base_name.json" \
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -239,7 +239,7 @@ pub fn object_to_shared_lib(object: &[u8], output_filename: &Path) -> Result<()>
             )
         }
     } else {
-        String::from("libcario_native_runtime.a")
+        String::from("libcairo_native_runtime.a")
     };
 
     let args: Vec<Cow<'static, str>> = {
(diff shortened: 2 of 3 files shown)


scripts/bench-hyperfine.sh
```diff
@@ -90,7 +90,7 @@ run_bench() {
         -o "$OUTPUT_DIR/$base_name-march-native" \
         >> /dev/stderr
 
-    CAIRO_NATIVE_RUNTIME_LIBDIR="$ROOT_DIR/target/release" hyperfine \
+    CAIRO_NATIVE_RUNTIME_LIBRARY="$ROOT_DIR/target/release/libcairo_native_runtime.a" hyperfine \
```
edg-l, line 93: It would be better to remove this env var from here in its entirely, because the CI has this library on another place, and we already set it, so its better to leave it to the user instead of forcing it here.

Before it didn't affect the CI because this env var is actually unused.

src/ffi.rs
```diff
@@ -219,6 +220,28 @@ pub fn object_to_shared_lib(object: &[u8], output_filename: &Path) -> Result<()>
...
+                .canonicalize()
+                .expect("Failed to cannonicalize path");
+            String::from(
+                absolute_path
+                    .to_str()
+                    .expect("Absolute path contains non-utf8 characters"),
+            )
+        }
+    } else {
+        String::from("libcario_native_runtime.a")
```
edg-l, line 242: There is a typo here: `libcario_native_runtime` should be `libcairo_native_runtime`

edg-l (changes requested):

src/ffi.rs
varex83: Fixed, can you please approve the workflow, please?

edg-l: > Fixed, can you please approve the workflow, please?

done

edg-l (approved):

igaray (approved):

PR #843, opened 2024-10-10
Better function attributes and re-enable >O1 opt

edg-l:
With these function attributes the dead code elimination can remove a lot of useless functions that aren't called after the inline pass has done its job.

This pr also sets the tests optlevel to Less so atleast some opts are done, so we can make sure the tests work even after llvm transformations.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -152,8 +152,8 @@ pub fn module_to_object(module: &Module<'_>, opt_level: OptLevel) -> Result<Vec<
         let opt = match opt_level {
             OptLevel::None => 0,
             OptLevel::Less => 1,
-            // slp-vectorizer pass may cause issues, but after the change
-            // on function attributes it seems to not trigger it anymore.
+            // slp-vectorizer pass did cause some issues, but after the change
+            // on function attributes it seems to not trigger them anymore.
             // https://github.com/llvm/llvm-project/issues/107198
             OptLevel::Default => 2,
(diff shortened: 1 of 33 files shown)


edg-l:
The tx that failed https://starkscan.co/tx/0x00b2867a82804fcf3c9f99c11f58ee417f03e1524f44a3a5693f4fc293005761 with the opt level >1 doesn't appear to fail now with this change either.

src/ffi.rs
```diff
@@ -152,8 +152,11 @@ pub fn module_to_object(module: &Module<'_>, opt_level: OptLevel) -> Result<Vec<
         let opt = match opt_level {
             OptLevel::None => 0,
             OptLevel::Less => 1,
-            OptLevel::Default => 1, // todo: change once slp-vectorizer pass is fixed on llvm
-            OptLevel::Aggressive => 1, // https://github.com/llvm/llvm-project/issues/107198
+            // slp-vectorizer pass may cause issues, but after the change
+            // on function attributes it seems to not trigger it anymore.
```
azteca1998, line 156:
```suggestion
            // slp-vectorizer pass did cause some issues, but after the change
            // on function attributes it seems to not trigger them anymore.
```

azteca1998 (dismissed):

azteca1998 (approved):

igaray (approved):

PR #849, opened 2024-10-14
fix aot contract executor not passing builtinstats

edg-l:
Found it executing some transactiosn after updating starknet-replay to latest

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -10,7 +10,7 @@ use crate::{
     execution_result::{BuiltinStats, ExecutionResult},
     starknet::{handler::StarknetSyscallHandlerCallbacks, StarknetSyscallHandler},
     types::TypeBuilder,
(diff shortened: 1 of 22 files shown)


JulianGCalderon (approved):

src/executor.rs
```diff
@@ -177,7 +177,6 @@ fn invoke_dynamic(
                     .to_bytes(&mut invoke_data)?;
             }
             CoreTypeConcrete::BuiltinCosts(_) => {
-                // This builtin should never be an argument but just in case.
```
azteca1998, line 180: Can it be an argument now?

azteca1998 (approved):

src/executor.rs
edg-l: yes, according to some tests with replay

Issue #845, opened 2024-10-10
[BUG] Native execution error: Wrong hash value


pefontana:
Running Starknet Replay with state_dump, native vs VM, there are some transactions that return the error
`Native execution error: Native execution error: Wrong hash value`, while in the VM executes OK

TX:
* block742069/0x5dfecafe396c3e26054cd34a56c52d6faf72e6c517a77f97aa2d042d3368f6c
* block742413/0x8c516e2682f2bb3f513636b7c746cb76e7fa7f8f521fa34609aa0fbbd5dee1
* block742697/0x7071b7f0cddda19cd5a66a6d3b422a33e2a92cc8c29b80d1b73ec1372bb517f
* block742924/0x638912a5f5bc27d30a6831ed14a677c1a4e7378a98f70af1827d47185f65c49.json
* block743548/0x1f0170875fd26726dfa57743556869b4f5adb94ccfecfd03ac87971b9ffb70
* block744354/0x377eebec36a704ac60e56c27a8f463b7862e60fb657567a2379f4bdbe575afa
* block744579/0x46065b0650515f6a57ff75f08c92d23095091e7dbd7a9f6e9dd98a86b714584
* block745590/0x45b62c05ba575d14be1ead99a9ad7f0d61763e39201be028033c06595883f36
* block747545/0x522a391120ab5861c00f8e8f5703c135fc0cc0db538af5855e646a87b965ebb
* block748162/0x4b31e308359d898a8c107d951e74c50483d7cd033d0e87362ad20bde5f2d220

### commits
Cairo Native: 82c25b377f50904e1253b614c8719388277530e6
Starknet replay: e125989afa77022995b3989c1d64292755312a87
blockifier = { git = "https://github.com/lambdaclass/sequencer", branch = "native2.8.x"},  1eaaa4b9902cd9f4a867b111ca87c3a3bfce5d29

PR #851, opened 2024-10-14
Fix process block bug

JulianGCalderon:
Fixes https://github.com/lambdaclass/cairo_native/issues/845, in conjunction with https://github.com/lambdaclass/sequencer/pull/11.

`wrap_sha256_process_block` syscall should modify the received state with the new state, but instead of doing this, it copies the value and modifies the copied version, leaving the old state untouched.

This PR fixes it.

All transactions mentioned in the issue are now working correctly.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/starknet.rs
+++ b/src/starknet.rs
@@ -1750,7 +1750,8 @@ pub(crate) mod handler {
             state: *mut [u32; 8],
             block: &[u32; 16],
         ) {
-            let result = ptr.sha256_process_block(unsafe { &mut *state }, block, gas);
+            let state_ref = unsafe { state.as_mut().unwrap() };
+            let result = ptr.sha256_process_block(state_ref, block, gas);
 
             *result_ptr = match result {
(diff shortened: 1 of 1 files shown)


src/starknet.rs
```diff
@@ -1750,7 +1750,8 @@ pub(crate) mod handler {
             state: *mut [u32; 8],
             block: &[u32; 16],
         ) {
-            let result = ptr.sha256_process_block(unsafe { &mut *state }, block, gas);
+            let state_ref = unsafe { state.as_mut().unwrap() };
+            let result = ptr.sha256_process_block(state_ref, block, gas);
```
azteca1998, line 1754: I think the following would be better:
```rs
// ...
            state: &mut [u32; 8],
            block: &[u32; 16],
        ) {
            let result = ptr.sha256_process_block(state, block, gas);
// ...
```

azteca1998 (approved):

PR #853, opened 2024-10-15
Also fix felt bits in starknet syscall wrappers

edg-l:
Related to https://github.com/lambdaclass/cairo_native/pull/844 but it missed some starknet syscalls

Refactored a bit the code to avoid some repetition.

This may fix some bugs we have, we need to test.

### DIFF
--- a/src/starknet.rs
+++ b/src/starknet.rs
@@ -16,7 +16,7 @@ pub struct ArrayAbi<T> {
 
 impl From<&ArrayAbi<Felt252Abi>> for Vec<Felt> {
     fn from(value: &ArrayAbi<Felt252Abi>) -> Self {
-        let felt_vec: Vec<_> = unsafe {
+        unsafe {
             let since_offset = value.since as usize;
             let until_offset = value.until as usize;
             debug_assert!(since_offset <= until_offset);
@@ -28,8 +28,7 @@ impl From<&ArrayAbi<Felt252Abi>> for Vec<Felt> {
         }
         .iter()
         .map(Felt::from)
-        .collect();
-        felt_vec
+        .collect()
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/starknet.rs
```diff
@@ -14,10 +14,45 @@ pub struct ArrayAbi<T> {
...
+            let len = until_offset - since_offset;
+            match len {
+                0 => &[],
+                _ => std::slice::from_raw_parts(value.ptr.add(since_offset), len),
+            }
+        }
+        .iter()
+        .map(Felt::from)
+        .collect();
+        felt_vec
```
azteca1998, line 32: There's no need to have a variable in this case.

azteca1998 (approved):

src/starknet.rs
```diff
@@ -1106,16 +1132,8 @@ pub(crate) mod handler {
...
             deploy_from_zero: bool,
         ) {
-            let class_hash = Felt::from_bytes_be(&{
-                let mut data = class_hash.0;
-                data.reverse();
-                data
-            });
-            let contract_address_salt = Felt::from_bytes_be(&{
-                let mut data = contract_address_salt.0;
-                data.reverse();
```
FrancoGiachetta, line 1116: There are some cases like this one in which the bytes are reversed before creating the felt, wouldn't it affect the result? It doesn't look like the `From<...>` implementation covers these cases, right?

edg-l: Notice that i use from_bytes_le in the from impl, and this snippet uses from_bytes_be, the reverse turns it into big endian, so it is no longer needed, this is because in the past there wasnt a from_bytes_le method so we had to do this

FrancoGiachetta: Ow right, I didn't catch that one.

FrancoGiachetta (approved):

azteca1998 (approved):

PR #869, opened 2024-10-18
Blocks Running Workflow

### LINKED ISSUE #852: Add blocks re execution in the CI
Similar to #850 but for every PR and with a smaller amount of blocks
We must add a workflow in Cairo Native that executes blocks with Starknet Replay with state_dump feature and compare the execution vs Cairo VM. To know if we are adding any new bug.
Probably, we should be able to select what Blockifier commit to use https://github.com/lambdaclass/starknet-replay/blob/main/Cargo.toml#L13
The workflow should not have any cache or previously compiled programs

FrancoGiachetta:
This PR adds a workflow which executes a small range of blocks both the vm and native, dumps the final state that the execution throws and compares with each other. If there's any diff, it will fail.

Closes #852 

## Checklist
- [X] Linked to Github Issue
- [ ] Unit tests added


.github/workflows/starknet-blocks.yml
```diff
@@ -0,0 +1,138 @@
...
+          path: replay
+
+      - name: Install Starknet Replay deps
+        run: make deps
+
+      - name: Patch Cairo Native's version
+        if: ${{ matrix.runner == 'native' }}
+        run: |
+          cd replay
+          sed -r -i 's/cairo_native.*rev.*[0-9a-z]+./cairo_native.git", rev = "${{ github.event.pull_request.head.sha }}"/' Cargo.toml
```
pefontana, line 91: nice one!

.github/workflows/starknet-blocks.yml
```diff
@@ -0,0 +1,138 @@
...
+          cargo update
```
pefontana, line 92: `cargo update` updates all the dependencies versions, in case the Cargo.toml allows it. I think it will be better to run a cargo build to ensure that the native change in the .toml is updated
`cargo build --release --bin replay`

scripts/diff-check.sh
pefontana, line 1: Here the rpc may fail and we skip certain state dumps compare.
I recoment seeing the script modification here https://github.com/lambdaclass/starknet-replay/pull/73/files file: scripts/cmp_state_dumps.sh
And also assert the skipped == 0

pefontana (changes requested):
Amaizing work @FrancoGiachetta !!!

FrancoGiachetta:
Changes done

JulianGCalderon (commented):
I think this works, but it's difficult to say as it can only be tested in the CI. Just to make sure, could you test it with some failing blocks? That way we can ensure that the workflow detects blocks with diffs.

FrancoGiachetta:
Yes, makes sense. I will replace some of them with the ones that seem to be failing recently

JulianGCalderon:
It seems to be working OK. Should we tag the workflow as `Required`?

pefontana:
> It seems to be working OK. Should we tag the workflow as `Required`?

Sure! I will add them

pefontana (approved):

JulianGCalderon (approved):

PR #882, opened 2024-10-25
Fix block workflow

FrancoGiachetta:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/.github/workflows/starknet-blocks.yml
+++ b/.github/workflows/starknet-blocks.yml
@@ -80,6 +80,12 @@ jobs:
       - name: Install Starknet Replay deps
         run: make deps
 
+      - name: Patch Cairo Native's version
+        if: ${{ matrix.runner == 'native' }}
+        run: |
+          cd replay
+          sed -r -i 's/cairo-native.*/cairo-native = { path = "..\/cairo_native" }/' Cargo.toml
+
       - name: Run Blocks Native
         if: ${{ matrix.runner == 'native' }}
         run: |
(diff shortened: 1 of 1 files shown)


edg-l (dismissed):

edg-l (dismissed):

.github/workflows/starknet-blocks.yml
```diff
@@ -81,12 +80,6 @@ jobs:
       - name: Install Starknet Replay deps
         run: make deps
 
-      - name: Patch Cairo Native's version
-        if: ${{ matrix.runner == 'native' }}
-        run: |
-          cd replay
-          sed -r -i 's/cairo_native.*rev.*[0-9a-z]+./cairo_native.git", rev = "${{ github.event.pull_request.head.sha }}"/' Cargo.toml
-
```
JulianGCalderon, line 89: This will only test against the default native commit on replay branch, right? We should patch de dependency to a local path, and use our current branch

JulianGCalderon (changes requested):

.github/workflows/starknet-blocks.yml
FrancoGiachetta: Ow, I thought is was already being done in the main branch, I'll change it

FrancoGiachetta: should work now.

edg-l (approved):

PR #888, opened 2024-10-25
Add daily block execution

### LINKED ISSUE #850: Add daily check re executing Starknet replay blocks
The idea is to add a workflow in Cairo Native that executes every day at the same hour.
The workflow should:
* Run several blocks with Starknet Replay with state_dump feature and compare the execution vs Cairo VM. If we find any discrepancies we should create an issue
* Starknet replay should use the last commit of Cairo Native main branch https://github.com/lambdaclass/starknet-replay/blob/main/Cargo.toml#L21. Probably we should be able to select what Blockifier commit to use https://github.com/lambdaclass/starknet-replay/blob/main/Cargo.toml#L13

Things to take into account:
* The workflow should not have any cache or previously compiled programs
* We should discuss what blocks are best to run, but for a first implementation could be any

JulianGCalderon:
Closes #850

You can find an example on how the workflow looks like [here](https://github.com/lambdaclass/cairo_native/actions/runs/11554827116)

It should work better once https://github.com/lambdaclass/starknet-replay/pull/78 is merged


.github/workflows/daily.yml
```diff
@@ -0,0 +1,95 @@
...
+        run: |
+          BLOCK_START=${{ matrix.block }}
+          BLOCK_END=$(($BLOCK_START + $RANGE_SIZE - 1))
+          cargo run --release --features state_dump block-range $BLOCK_START $BLOCK_END mainnet
+        continue-on-error: true
+      - name: Run with VM
+        run: |
+          BLOCK_START=${{ matrix.block }}
+          BLOCK_END=$(($BLOCK_START + $RANGE_SIZE - 1))
+          cargo run --release --features state_dump,only_cairo_vm block-range $BLOCK_START $BLOCK_END mainnet
```
FrancoGiachetta, line 91: Since this is meant to be run during night, I don't really know if this could make a significant impact on the time it takes to complete the ci once we increased the amount of blocks to execute. But it might be a good idea to run both native and the vm at the same time to reduce the execution time, right?.

FrancoGiachetta (approved):

edg-l (approved):

.github/workflows/daily.yml
JulianGCalderon: I considered it, but I think is not that simple. As I understand it, we have two possible approaches:

The first approach (used in the `starknet-blocks` workflow) is to have two separate jobs, one uses a matrix to execute different blocks with native and VM, then uploads an artifact with the state dumps. Another job depends on the first one, retreives the artifact, and compares the states. 

The downside with this approach is that if one of the executions fails, then the `compare` job is never ran. This is not desirable, as an RPC error is pretty common when running so many blocks, and would cancel the whole daily execution. There might be a way to work around this, but I don't know how.

This is why I opted for a different approach. Use a matrix, but each job in the matrix is complete and independent from the other ones. A failure in one execution would not cancel the other ones, as they are completely separate. The comparison would not have to wait for the whole execution, as each subcomparison could start as soon as the corresponding execution finishes.

The downside of this is that we cannot run VM and Native at the same time, but I don't think that it matters to much, as we can always increase the amount of jobs in the matrix if we want more parallelism.

PR #891, opened 2024-10-30
Fix daily block run

JulianGCalderon:
Run Native and VM in parallel, and creates issues on failing runs

An example issue can be found [here](https://github.com/lambdaclass/cairo_native/issues/894)


.github/workflows/daily.yml
```diff
@@ -32,12 +31,17 @@ jobs:
...
+        runner:
+          - native
+          - vm
       fail-fast: false
     defaults:
       run:
         shell: bash
         working-directory: ./starknet-replay
 
+    # todo: build first on a single job, and run the blocks after
```
pefontana, line 43: Should this work?
I think that since there are different workflows, we can build all just once

JulianGCalderon: I saw that in the VM something similar is done: https://github.com/lambdaclass/cairo-vm/blob/159f67da19964cc54a95423a69470a26e534a13d/.github/workflows/rust.yml#L393-L410. I thought it would be nice to add it, although in a 6h workflow, building the binary may be negligible, so maybe is not that worth it

pefontana: ahhh okay, if we save it in cache it could work, but as you said I think it won`t make a significant change
let just remove this todo

JulianGCalderon: agreed, I'll remove it

pefontana (approved):

edg-l (approved):

PR #897, opened 2024-11-04
Make arrays shared cow v2

azteca1998:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/libfuncs/array.rs
```diff
@@ -1098,128 +1545,292 @@ pub fn build_slice<'ctx, 'this>(
...
         location,
     ));
 
     {
-        let elem_ty = elem_ty.build(context, helper, registry, metadata, &info.ty)?;
-
-        let value = entry.arg(1)?;
-        let value = slice_block.insert_value(context, location, value, slice_start, 1)?;
-        let value = slice_block.insert_value(context, location, value, slice_end, 2)?;
+        // TODO: If shared -> Clone and drop.
```
edg-l, line 1610: remove comment if done

edg-l (approved):

edg-l (approved):

edg-l (dismissed):

src/types/array.rs
```diff
@@ -14,6 +14,11 @@
 //! |   1   | `i32`          | Array end offset[^2].    |
 //! |   2   | `i32`          | Allocated capacity[^2].  |
 //!
+//! The pointer to the allocation (which is **not the data**) contains:
+//!   1. Reference counter.
+//!   2. Padding.
+//!   3. Array data. Its address is the pointer to the data stored in the type.
+//!
```
JulianGCalderon, line 21: I'm not sure I understood it correctly. What do you mean by _allocation_? 

If the index 0 points to the array data, then to calculate the counter, you need to substract (padding + counter size) from the data pointer?

azteca1998: The call to `malloc` returns a pointer, which is the allocation pointer; aka, the pointer pointing to the allocation. This pointer is not what is stored within the array struct.

The allocation pointer has:
  - The reference counter (i32).
  - Padding to align the data.
  - The actual data (the pointer to which is stored in the array struct).

Previously the allocation pointer was the same as the data pointer, but this has been changed. Now, to free the array we need to subtract the offset calculated by the reference counter and the padding. This has been made because most of the time we don't need the reference counter and having to offset the pointer every time would add boilerplate and unnecessary operations.

> If the index 0 points to the array data, then to calculate the counter, you need to substract (padding + counter size) from the data pointer?

Yes.

src/libfuncs/array.rs
```diff
@@ -1237,309 +1846,137 @@ pub fn build_slice<'ctx, 'this>(
     Ok(())
 }
 
-/// Generate MLIR operations for the `span_from_tuple` libfunc.
-pub fn build_span_from_tuple<'ctx, 'this>(
+/// Generate MLIR operations for the `array_len` libfunc.
+pub fn build_len<'ctx, 'this>(
     context: &'ctx Context,
-    registry: &ProgramRegistry<CoreType, CoreLibfunc>,
+    _registry: &ProgramRegistry<CoreType, CoreLibfunc>,
```
gabrielbosio, line 1852: Why we need this unused argument? Can we remove it?

examples/starknet.rs
```diff
@@ -43,6 +45,7 @@ struct TestingState {
 }
 
 #[derive(Debug, Default)]
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
azteca1998: They are not strictly necessary, but they make a few warnings disappear when testing.
The `#[allow(dead_code)]` directives are only in the `examples/` folder and not within the actual codebase.

src/libfuncs/array.rs
azteca1998: Yes, but I'd rather leave it for consistency. All libfunc implementations have the same arguments regardless of whether they're being used or not.

It's convenient when we have to fix something and all the data we need is already available, even if we weren't using it before.

src/libfuncs/array.rs
```diff
@@ -371,720 +423,1115 @@ pub fn build_get<'ctx, 'this>(
...
+        context,
+        CmpiPredicate::Eq,
+        array_capacity,
+        k0,
+        location,
+    ))?;
 
-        // There's no need to clone it since we're moving it out of the array.
-        valid_block.memcpy(context, location, elem_ptr, target_ptr, elem_size);
+    let is_shared = entry.append_op_result(scf::r#if(
```
JulianGCalderon, line 521: Can you reuse the `is_shared` function here?

src/libfuncs/array.rs
```diff
@@ -371,720 +423,1115 @@ pub fn build_get<'ctx, 'this>(
...
+    Single(&'a SignatureAndTypeConcreteLibfunc),
+    Multi(&'a ConcreteMultiPopLibfunc),
+}
+
+/// Generate MLIR operations for the `array_pop_*` libfuncs.
+///
+/// Template arguments:
+///   - Consume: Whether to consume or not the array on failure.
+///   - Reverse: False for front-popping, true for back-popping.
+///   - Multiple: True for multi-popping.
```
JulianGCalderon, line 878: This comment may be outdated, as there is no `MULTIPLE` const argument.

src/libfuncs/array.rs
JulianGCalderon: Aah, you are already doing it. You can ignore this comment.

src/libfuncs/array.rs
```diff
@@ -371,720 +423,1115 @@ pub fn build_get<'ctx, 'this>(
...
 
-    let value = entry.arg(0)?;
+                        block.gep(
+                            context,
+                            location,
+                            array_ptr,
+                            &[GepIndex::Value(offset)],
+                            IntegerType::new(context, 8).into(),
+                        )?
+                    };
```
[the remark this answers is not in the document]
azteca1998: - `array_ptr`: The pointer that is stored in the `!llvm.struct<(ptr, i32, i32, i32)>`. It may or may not point to the data.
  - `data_ptr`: It's `array_ptr + elem_stride * array_start`. It always points to the start of the data. It'll be equivalent to `array_ptr` when `array_start` is zero.

When popping in reverse we start copying data from where the data start and skip the popped elements, thus the start pointer is `data_ptr`. The offset in the cloned array will always be zero (since we're copying the data anyways we'll also defragment it.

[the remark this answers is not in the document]
JulianGCalderon: Aah, I get it now. Thanks

JulianGCalderon (approved):

edg-l (approved):

PR #904, opened 2024-11-12
Remove panic in metadata

FrancoGiachetta:
- [x] src/metadata.rs
- [x] src/metadata/auto_breakpoint.rs
- [x] src/metadata/drop_overrides.rs
- [x] src/metadata/dup_overrides.rs
- [x] src/metadata/gas.rs
- [x] src/metadata/realloc_bindings.rs
- [x] src/metadata/runtime_bindings.rs

There are some other files not mentioned in the list also with some changes (simply adding the `?` keyword where methods returning `Result` are used), but those are due to a change in the free's and alloc's signature (`src/metadata/realloc_bindings.rs`). Which does not mean their unwraps were also removed.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -63,7 +63,7 @@ impl MetadataStorage {
     {
         self.entries.remove(&TypeId::of::<T>()).map(|meta| {
             *(Box::<(dyn Any + 'static)>::downcast::<T>(meta)
-                .)
+                .expect("the given type does not match the actual"))
         })
     }
 
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -63,7 +63,7 @@ impl MetadataStorage {
     {
         self.entries.remove(&TypeId::of::<T>()).map(|meta| {
             *(Box::<(dyn Any + 'static)>::downcast::<T>(meta)
-                .expect("the given type does not match the actual"))
+                .)
         })
     }
 
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -62,8 +62,9 @@ impl MetadataStorage {
         T: Any,
     {
         self.entries.remove(&TypeId::of::<T>()).map(|meta| {
-            *(Box::<(dyn Any + 'static)>::downcast::<T>(meta)
-                .expect("the given type does not match the actual"))
+            *(Box::<(dyn Any + 'static)>::downcast::<T>(meta).expect(
+                "attempt to downcast a boxed value to a type which does not match the actual",
+            ))
         })
     }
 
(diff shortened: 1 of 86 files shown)


src/metadata.rs
```diff
@@ -61,9 +61,10 @@ impl MetadataStorage {
     where
         T: Any,
     {
-        self.entries
-            .remove(&TypeId::of::<T>())
-            .map(|meta| *(Box::<(dyn Any + 'static)>::downcast::<T>(meta).unwrap()))
+        self.entries.remove(&TypeId::of::<T>()).map(|meta| {
+            *(Box::<(dyn Any + 'static)>::downcast::<T>(meta)
+                .expect("the given type does not match the actual"))
+        })
```
JulianGCalderon, line 67: Shouldn't we also remove the expects in this file?

src/metadata/gas.rs
```diff
@@ -100,17 +102,19 @@ impl GasMetadata {
         Some(
             self.gas_info.function_costs[func]
                 .iter()
-                .map(|(token_type, val)| val.into_or_panic::<usize>() * token_gas_cost(*token_type))
+                .map(|(token_type, val)| (*val as usize) * token_gas_cost(*token_type))
```
JulianGCalderon, line 105: Using `as` can also fail, but it will do so silently. I think it's better to return an error in those cases (a NativeAssert)

JulianGCalderon (changes requested):

src/metadata.rs
FrancoGiachetta: Yes, I did it that way when native_assert still wasn't there, and I totally forgot of changing it. Thanks, I'll change it

src/metadata/gas.rs
FrancoGiachetta: The problem with that is that the `initial_required_gas` is used by `get_initial_available_gas` which already returns an error. So I don't think that returning a NativeAssert would solve this situation.

src/metadata.rs
FrancoGiachetta: Gave it a second thought, we are having the same issue as in libfuncs. The error might happen inside the map, so the only solution to remove the panic would be to make it eager.

JulianGCalderon (approved):

azteca1998 (approved):

edg-l (approved):

JulianGCalderon (approved):

edg-l (approved):

PR #905, opened 2024-11-12
Remove panic in compiler.rs

JulianGCalderon:
Adds a new error type: `NativeAssertError`. It's a replacement for `panic!`, as it collects the backtrace but it doesn't interrupt the flow of the program.

There is a single `todo!` left, but I didn't know how to deal with it. We should take a closer look at it.

There are a few `expect` left, but being such a simple case, I'm not sure is worth handling it.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
- [ ] Integration tests added.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.

PR #906, opened 2024-11-12
Remove panic from compiler.rs (2)

JulianGCalderon:
Adds a new error type: `NativeAssertError`. It's a replacement for panic!, as it collects the backtrace but it doesn't interrupt the flow of the program.

There is a single `todo!` left, but I didn't know how to deal with it. We should take a closer look at it.

There are a few `expect` left, but being such a simple case, I'm not sure is worth handling it.

Unlike #905, I tried to make this non-panic abstraction as similar as possible to the associated panicking code. The compiler.rs was changed to use the non-panic variants

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -46,13 +46,14 @@
 
 use crate::{
     debug::libfunc_to_name,
-    error::{Error, NativeAssertError, NativeAssertErrorKind},
+    error::{Error, ToNativeExpect},
     libfuncs::{BranchArg, LibfuncBuilder, LibfuncHelper},
     metadata::{
         gas::{GasCost, GasMetadata},
         tail_recursion::TailRecursionMeta,
         MetadataStorage,
     },
+    native_panic,
     types::TypeBuilder,
     utils::{generate_function_name, BlockExt},
 };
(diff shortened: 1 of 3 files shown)


src/compiler.rs
```diff
@@ -1367,11 +1373,13 @@ fn generate_entry_point_wrapper<'c>(
...
                 Identifier::new(context, "llvm.linkage"),
-                Attribute::parse(context, "#llvm.linkage<private>").unwrap(),
+                Attribute::parse(context, "#llvm.linkage<private>")
+                    .expect("hardcoded attribute is valid"),
             ),
             (
                 Identifier::new(context, "llvm.CConv"),
-                Attribute::parse(context, "#llvm.cconv<fastcc>").unwrap(),
+                Attribute::parse(context, "#llvm.cconv<fastcc>")
+                    .expect("hardcoded attribute is valid"),
```
FrancoGiachetta, line 1382: shouldn't it be "invalid" instead of "valid"?

JulianGCalderon: You are right. FIxed

JulianGCalderon: I changed the code to use that error variant. Thanks for pointing it out.

FrancoGiachetta (approved):
LGTM

edg-l (approved):

PR #907, opened 2024-11-12
Remove panic in context.rs

JulianGCalderon:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
- [ ] Integration tests added.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.

PR #917, opened 2024-11-13
Remove panics in libfuncs

FrancoGiachetta:
- [x] src/libfuncs.rs
- [x] src/libfuncs/array.rs
- [x] src/libfuncs/bool.rs
- [x] src/libfuncs/cast.rs
- [x] src/libfuncs/const.rs
- [x] src/libfuncs/debug.rs
- [x] src/libfuncs/enum.rs
- [x] src/libfuncs/felt252.rs
- [x] src/libfuncs/felt252_dict.rs
- [x] src/libfuncs/felt252_dict_entry.rs
- [x] src/libfuncs/function_call.rs
- [x] src/libfuncs/gas.rs
- [x] src/libfuncs/pedersen.rs
- [x] src/libfuncs/poseidon.rs
- [x] src/libfuncs/starknet.rs
- [x] src/libfuncs/starknet/testing.rs

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs.rs
+++ b/src/libfuncs.rs
@@ -2,7 +2,11 @@
 //!
 //! Contains libfunc generation stuff (aka. the actual instructions).
 
-use crate::{error::Error as CoreLibfuncBuilderError, metadata::MetadataStorage, utils::BlockExt};
+use crate::{
+    error::{panic::ToNativeAssertError, Error as CoreLibfuncBuilderError, Result as NativeResult},
+    metadata::MetadataStorage,
+    utils::BlockExt,
+};
 use bumpalo::Bump;
 use cairo_lang_sierra::{
     extensions::core::{CoreConcreteLibfunc, CoreLibfunc, CoreType},
@@ -270,14 +274,14 @@ where
     'this: 'ctx,
 {
     #[doc(hidden)]
-    pub(crate) fn results(self) -> impl Iterator<Item = Vec<Value<'ctx, 'this>>> {
+    pub(crate) fn results(self) -> impl Iterator<Item = NativeResult<Vec<Value<'ctx, 'this>>>> {
         self.results.into_iter().enumerate().map(|(branch_idx, x)| {
             x.into_iter()
                 .enumerate()
                 .map(|(arg_idx, x)| {
-                    x.into_inner().unwrap_or_else(|| {
-                        panic!("Argument #{arg_idx} of branch {branch_idx} doesn't have a value.")
-                    })
+                    x.into_inner().to_native_assert_error(
+                        "Argument #{arg_idx} of branch {branch_idx} doesn't have a value.",
+                    )
                 })
                 .collect()
         })
--- a/src/libfuncs.rs
+++ b/src/libfuncs.rs
@@ -279,9 +279,9 @@ where
             x.into_iter()
                 .enumerate()
                 .map(|(arg_idx, x)| {
-                    x.into_inner().to_native_assert_error(
-                       &format!("Argument #{arg_idx} of branch {branch_idx} doesn't have a value."),
-                    )
+                    x.into_inner().to_native_assert_error(&format!(
+                        "Argument #{arg_idx} of branch {branch_idx} doesn't have a value."
+                    ))
                 })
                 .collect()
         })
--- a/src/libfuncs.rs
+++ b/src/libfuncs.rs
@@ -275,12 +275,12 @@ where
 {
     #[doc(hidden)]
     pub(crate) fn results(self) -> impl Iterator<Item = NativeResult<Vec<Value<'ctx, 'this>>>> {
-        self.results.into_iter().enumerate().map(|(_, x)| {
+        self.results.into_iter().enumerate().map(|(branch_idx, x)| {
             x.into_iter()
                 .enumerate()
-                .map(|(_, x)| {
+                .map(|(arg_idx, x)| {
                     x.into_inner().to_native_assert_error(
-                        "Argument #{arg_idx} of branch {branch_idx} doesn't have a value.",
+                       &format!("Argument #{arg_idx} of branch {branch_idx} doesn't have a value."),
                     )
                 })
                 .collect()
--- a/src/libfuncs.rs
+++ b/src/libfuncs.rs
@@ -275,10 +275,10 @@ where
 {
     #[doc(hidden)]
     pub(crate) fn results(self) -> impl Iterator<Item = NativeResult<Vec<Value<'ctx, 'this>>>> {
-        self.results.into_iter().enumerate().map(|(branch_idx, x)| {
+        self.results.into_iter().enumerate().map(|(_, x)| {
             x.into_iter()
                 .enumerate()
-                .map(|(arg_idx, x)| {
+                .map(|(_, x)| {
                     x.into_inner().to_native_assert_error(
                         "Argument #{arg_idx} of branch {branch_idx} doesn't have a value.",
                     )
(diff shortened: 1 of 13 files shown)


src/libfuncs.rs
```diff
@@ -270,14 +274,14 @@ where
...
     #[doc(hidden)]
-    pub(crate) fn results(self) -> impl Iterator<Item = Vec<Value<'ctx, 'this>>> {
-        self.results.into_iter().enumerate().map(|(branch_idx, x)| {
+    pub(crate) fn results(self) -> impl Iterator<Item = NativeResult<Vec<Value<'ctx, 'this>>>> {
+        self.results.into_iter().enumerate().map(|(_, x)| {
             x.into_iter()
                 .enumerate()
-                .map(|(arg_idx, x)| {
-                    x.into_inner().unwrap_or_else(|| {
-                        panic!("Argument #{arg_idx} of branch {branch_idx} doesn't have a value.")
```
edg-l, line 279: i think you are not writing the arg_idx value into the string here

FrancoGiachetta: Ow you are right, thats why it wasn't being use, changing it, thanks.

FrancoGiachetta: It should be right now

src/libfuncs.rs
```diff
@@ -270,14 +274,14 @@ where
     'this: 'ctx,
 {
     #[doc(hidden)]
-    pub(crate) fn results(self) -> impl Iterator<Item = Vec<Value<'ctx, 'this>>> {
+    pub(crate) fn results(self) -> impl Iterator<Item = NativeResult<Vec<Value<'ctx, 'this>>>> {
```
azteca1998, line 277:
```rs
pub(crate) fn results(self) -> impl Iterator<Item = NativeResult<Vec<Value<'ctx, 'this>>>> {
```
If you're going to do this, then returning `Result<impl Iterator<...>>` would be better. Like this:
```rs
pub(crate) fn results(self) -> NativeResult<impl Iterator<Item = Vec<Value<'ctx, 'this>>>> {
    // ...
}
```

src/libfuncs/bounded_int.rs
```diff
@@ -114,8 +115,13 @@ fn build_add<'ctx, 'this>(
     let compute_ty = IntegerType::new(context, compute_range.offset_bit_width()).into();
 
     // Zero-extend operands into the computation range.
-    assert!(compute_range.offset_bit_width() >= lhs_width);
-    assert!(compute_range.offset_bit_width() >= rhs_width);
+    if !compute_range.offset_bit_width() >= lhs_width {
+        native_panic!("the lhs_range bit_width must be less or equal than the compute_range");
+    }
```
azteca1998, line 120: Please make a `native_assert` macro instead of this (in all instances).

src/libfuncs/enum.rs
```diff
@@ -119,11 +120,13 @@ pub fn build_enum_value<'ctx, 'this>(
         helper,
         registry,
         metadata,
-        type_info.variants().unwrap(),
+        type_info
+            .variants()
+            .to_native_assert_error("couldn't get enum's variants")?,
```
azteca1998, line 125:
```suggestion
            .to_native_assert_error("found non-enum type where an enum is required")?,
```
It's not that we can't get an enum's variants because if the type is an enum then they'll always be available. It's rather that the type is not an enum therefore it has no variants.

src/libfuncs/enum.rs
```diff
@@ -257,7 +260,9 @@ pub fn build_match<'ctx, 'this>(
 ) -> Result<()> {
     let type_info = registry.get_type(&info.param_signatures()[0].ty)?;
 
-    let variant_ids = type_info.variants().unwrap();
+    let variant_ids = type_info
+        .variants()
+        .to_native_assert_error("couldn't get enum's variants")?;
```
azteca1998, line 265: Same here.

azteca1998 (changes requested):

src/libfuncs/array.rs
```diff
@@ -1485,9 +1484,9 @@ pub fn build_tuple_from_span<'ctx, 'this>(
         CoreTypeConcrete::Array(info) => (&info.ty, registry.get_type(&info.ty)?),
         CoreTypeConcrete::Snapshot(info) => match registry.get_type(&info.ty)? {
             CoreTypeConcrete::Array(info) => (&info.ty, registry.get_type(&info.ty)?),
-            _ => unreachable!(),
+            _ => native_panic!("unexpected CoreTypeConcrete found"),
         },
-        _ => unreachable!(),
+        _ => native_panic!("unexpected CoreTypeConcrete found"),
```
JulianGCalderon, line 1489: Could you give a little more information on why it's an unexpected type?

src/libfuncs/array.rs
```diff
@@ -1508,8 +1507,14 @@ pub fn build_tuple_from_span<'ctx, 'this>(
 
     let array_len = entry.append_op_result(arith::subi(array_end, array_start, location))?;
     let (tuple_len, tuple_len_val) = {
-        let fields = registry.get_type(&info.ty)?.fields().unwrap();
-        assert!(fields.iter().all(|f| f.id == elem_id.id));
+        let fields = registry
+            .get_type(&info.ty)?
+            .fields()
+            .to_native_assert_error("missing filed")?;
```
JulianGCalderon, line 1513: typo
```suggestion
            .to_native_assert_error("missing field")?;
```

src/libfuncs/array.rs
```diff
@@ -1648,11 +1653,9 @@ pub fn build_tuple_from_span<'ctx, 'this>(
             &info.signature.param_signatures[0].ty,
         )?;
 
-        // The following unwrap is unreachable because an array always has a drop implementation,
-        // which at this point is always inserted thanks to the `build_type()` just above.
```
JulianGCalderon, line 1652: This comment gives valuable information on why the following operation will never fail. I think you should keep it, even though you now replaced the unwrap for a proper error. You could rephrase it like this: "the following operation will never fail because..."

src/libfuncs/bounded_int.rs
JulianGCalderon: I added an implementation of native_assert on [this PR.](https://github.com/lambdaclass/cairo_native/pull/907) You could take it from there.

src/libfuncs.rs
FrancoGiachetta: Yes, I first tried to tackle it that way. The problem is that the error might happen inside the `map()` and since the iterator expects not a `Result<Vec<Value<'ctx, 'this>>` but a ` Vec<Value<'ctx, 'this>`, it breaks.

src/libfuncs/bounded_int.rs
FrancoGiachetta: Done!

src/libfuncs/enum.rs
FrancoGiachetta: Done!

src/libfuncs/enum.rs
FrancoGiachetta: Done!

src/libfuncs.rs
JulianGCalderon: Aah, that's true. As long as the iterator is lazy we won't be able to move up the result. Could we make it eager? Usually libfuncs don't have that many result branches, right?

src/libfuncs/array.rs
FrancoGiachetta: Done!

src/libfuncs/array.rs
FrancoGiachetta: Done!

azteca1998 (approved):

src/libfuncs.rs
azteca1998: Yes, please make it eager.

edg-l (approved):

PR #919, opened 2024-11-13
Coalesce all compatible no-op libfuncs into a single function.

azteca1998:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/libfuncs.rs
```diff
@@ -435,3 +445,39 @@ fn increment_builtin_counter_by<'ctx: 'a, 'a>(
...
+    param_signatures: &[ParamSignature],
+) -> Result<()> {
+    let mut params = Vec::with_capacity(N);
+
+    #[allow(clippy::needless_range_loop)]
+    for i in 0..N {
+        let param_ty = registry.get_type(&param_signatures[i].ty)?;
+        let mut param_val = entry.argument(i)?.into();
+
+        if PROCESS_BUILTINS
```
JulianGCalderon, line 465: Couldn't we iterate over param_signatures instead of using a const generic?

src/libfuncs/mem.rs
```diff
@@ -35,39 +35,36 @@ pub fn build<'ctx, 'this>(
...
+        | MemConcreteLibfunc::StoreTemp(SignatureAndTypeConcreteLibfunc { signature, .. }) => {
+            super::build_noop::<1, false>(
+                context,
+                registry,
+                entry,
+                location,
+                helper,
+                metadata,
+                &signature.param_signatures,
+            )
```
JulianGCalderon, line 63: Out of curiosity, why some noops must not update builtin counters?

azteca1998: Libfuncs that receive builtins but do not use them as a builtin, for example `store_temp<Bitwise>`.

src/libfuncs.rs
azteca1998: While we technically could, in this case we depend on the constant `N`. It should never happen, but if the number of arguments was smaller than the constant then the number of values to return would be smaller than the constant `N`.

JulianGCalderon: Aaah. Is there a situation where the number of return values is smaller than the number of arguments?

azteca1998: None. This can only happen on:
  - Dev error (by typing the wrong constant). Should be caught by any test that uses the libfunc.
  - Invalid Sierra type info. This should also be impossible at least without breaking changes and only between versions. If I'm not mistaken the type info is hardcoded into the Sierra type enums.

JulianGCalderon: I'm not sure I understand the problem then. If we remove `const N: usize` and always iterate `param_signatures`, then we would never encounter a SIGSEGV and we will always return the same number of values as we receive, making any dev error impossible. Maybe I'm missing something. Is `const N` used for performance reasons?

JulianGCalderon (approved):

edg-l (dismissed):

src/libfuncs.rs
JulianGCalderon: @azteca1998 could you add a todo comment, reminding us to check if all noops return as many arguments as they receive?

azteca1998: No need, they do not. Take the guarantee verify libfuncs, for example:
```cairo
extern fn u128_mul_guarantee_verify(guarantee: U128MulGuarantee) implicits(RangeCheck) nopanic;
```
Their signature is as follows:
```
(RangeCheck, U128MulGuarantee) -> (RangeCheck);
```

JulianGCalderon: Aah, that is true.

JulianGCalderon (dismissed):

edg-l (approved):

JulianGCalderon (approved):

PR #924, opened 2024-11-14
Remove panic in utils.rs

JulianGCalderon:
There are some unwraps left, but they are only used in tests.

`get_integer_layout` will never fail, and adding error handling implies changing a lot of files, as it's used throughout the codebase. We can consider changing it later.

The file `utils/mem_tracing.rs` also contains some unreachable, but as it's a debugging feature I don't think it's worth to update it.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -26,7 +26,6 @@ use std::{
     alloc::Layout,
     borrow::Cow,
     fmt::{self, Display},
-    ops::Neg,
     path::Path,
 };
 use thiserror::Error;
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -233,9 +233,10 @@ pub fn felt252_str(value: &str) -> Felt {
     let value = value
         .parse::<BigInt>()
         .expect("value must be a digit number");
-    let value = match value.sign() {
-        Sign::Minus => &*PRIME - value.neg().to_biguint().unwrap(),
-        _ => value.to_biguint().unwrap(),
+
+    let value = match value.to_biguint() {
+        Some(value) => value,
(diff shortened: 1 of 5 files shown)


src/utils.rs
```diff
@@ -234,9 +233,10 @@ pub fn felt252_str(value: &str) -> Felt {
     let value = value
         .parse::<BigInt>()
         .expect("value must be a digit number");
-    let value = match value.sign() {
-        Sign::Minus => &*PRIME - value.neg().to_biguint().unwrap(),
-        _ => value.to_biguint().unwrap(),
+
+    let value = match value.to_biguint() {
+        Some(value) => value,
+        None => &*PRIME - value.neg().to_biguint().expect("is always positive"),
```
azteca1998, line 239: This unwrap can be removed altogether. Check out [this snippet](https://github.com/lambdaclass/cairo_native/blob/main/src/libfuncs/felt252.rs#L328-L333) for more info.

azteca1998 (dismissed):

src/utils.rs
JulianGCalderon: Aah, I couldn't figure out how to remove that unwrap. Thanks. This would also work, right?:
```rust
    let value = match value.sign() {
        Sign::Minus => &*PRIME - value.magnitude(),
        _ => value.magnitude().clone(),
    };
```

FrancoGiachetta (dismissed):
LGTM

edg-l (approved):

azteca1998 (approved):

PR #946, opened 2024-11-27
Implement libfunc profiler.

azteca1998:
Implementation of a libfunc profiler.

For each executed libfunc, generates:
  - Number of samples.
  - Total number of ticks.
  - Average number of ticks.
  - Standard deviation.
  - Quartiles (0%, 25%, 50%, 75%, 100%).

Also implements an outlier detection system to filter out operating system noise.

> Note: The `x86_64` architecture handles context switches between cores correctly, but in `aarch64` this is impossible. Results may not be reliable in `aarch64`.

Output Example:

```md
[2]
    Total Samples:          281
    Total Execution Time:   111
    Average Execution Time: 0.39501779359430605
    Standard Deviation:     0.23897873633819258
    Quartiles:              [0, 0, 0, 1, 1]

```
Where `[2]` refers to the libfunc id.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/bin/cairo-native-run.rs
```diff
@@ -46,6 +46,11 @@ struct Args {
     #[arg(short = 'O', long, default_value_t = 0)]
     opt_level: u8,
 
+    #[cfg(feature = "with-libfunc-profiling")]
+    #[arg(long)]
+    /// The output path for the libfunc profilling results
```
gabrielbosio, line 51:
```suggestion
    /// The output path for the libfunc profiling results
```

src/bin/cairo-native-run.rs
```diff
@@ -188,6 +214,31 @@ fn main() -> anyhow::Result<()> {
...
+
+        for (libfunc_id, (n_samples, sum, quartiles, average, std_dev)) in profile.process() {
+            writeln!(output, "{libfunc_id}")?;
+            writeln!(output, "    Samples  : {n_samples}")?;
+            writeln!(output, "    Sum      : {sum}")?;
+            writeln!(output, "    Average  : {average}")?;
+            writeln!(output, "    Deviation: {std_dev}")?;
+            writeln!(output, "    Quartiles: {quartiles:?}")?;
+            writeln!(output)?;
+        }
```
JulianGCalderon, line 239: Could you write the output of an example run in the description of the PR?

src/metadata/profiler.rs
```diff
@@ -0,0 +1,417 @@
...
+        let value_hi = block.extract_value(context, location, value, i32_ty, 0)?;
+        let value_lo = block.extract_value(context, location, value, i32_ty, 1)?;
+        let core_idx = block.extract_value(context, location, value, i32_ty, 2)?;
+
+        let value_hi = block.extui(value_hi, i64_ty, location)?;
+        let value_lo = block.extui(value_lo, i64_ty, location)?;
+        let value = block.shli(value_hi, k32, location)?;
+        let value = block.append_op_result(arith::ori(value, value_lo, location))?;
+
+        Ok((value, core_idx))
```
JulianGCalderon, line 196: What is `core_idx`? Could you add documentation on the return values?

src/metadata/profiler.rs
```diff
@@ -0,0 +1,417 @@
...
+
+        let Some(profiler) = profiler.get_mut(&trace_id) else {
+            eprintln!("Could not find libfunc profiler!");
+            return;
+        };
+
+        profiler
+            .trace
+            .push((StatementIdx(statement_idx as usize), tick_delta));
+    }
```
JulianGCalderon, line 318: Could you document that u64::MAX implies an invalid time delta?

src/libfuncs.rs
```diff
@@ -349,7 +357,16 @@ where
...
-        cf::br(successor, &destination_operands, location)
+        #[cfg(feature = "with-libfunc-profiling")]
+        self.push_profiler_frame(
+            unsafe { self.context().to_ref() },
+            self.module,
+            block,
+            location,
+        )?;
+
+        block.append_operation(cf::br(successor, &destination_operands, location));
```
JulianGCalderon, line 368: Why do we now append the operation, rather than just returning the `cf::br` operation? Idem for `cond_br`.

src/compiler.rs
```diff
@@ -620,6 +600,45 @@ fn compile_func(
...
+                        results: invocation
+                            .branches
+                            .iter()
+                            .map(|x| vec![Cell::new(None); x.results.len()])
+                            .collect::<Vec<_>>(),
+
+                        #[cfg(feature = "with-libfunc-profiling")]
+                        profiler: match libfunc {
+                            CoreConcreteLibfunc::FunctionCall(_) => {
+                                // Tail-recursive function calls are broken. Also it'd include the entire function which
```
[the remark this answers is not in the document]
FrancoGiachetta: Added better docs here

src/libfuncs.rs
FrancoGiachetta: We concluded this would be quite difficult to accomplish.

src/bin/cairo-native-run.rs
FrancoGiachetta: Done

src/metadata/profiler.rs
```diff
@@ -0,0 +1,417 @@
...
+        let value_lo = block.extui(value_lo, i64_ty, location)?;
+        let value = block.shli(value_hi, k32, location)?;
+        let value = block.append_op_result(arith::ori(value, value_lo, location))?;
+
+        Ok((value, core_idx))
+    }
+
+    /// Get the timestamp
+    #[cfg(target_arch = "aarch64")]
+    pub fn measure_timestamp<'c, 'a>(
```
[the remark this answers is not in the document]
FrancoGiachetta: It seems this is not as inaccurate as we thought, so perhaps it is safe to leave it as it is.

src/metadata/profiler.rs
```diff
@@ -0,0 +1,431 @@
...
+
+                // A tick_delta equal to u64::MAX implies it is invalid, so we don't take it
+                // into account
+                if *tick_delta != u64::MAX {
+                    tick_deltas.push(*tick_delta);
+                } else {
+                    *extra_count += 1;
+                }
+            }
+        }
```
JulianGCalderon, line 345: I think its better to separate the profile processing, from profile capturing.

We could move this method to a standalone function that receives:
1. The raw profile data
2. The sierra program.

And builds the sierra profile summary. I'd also move the definition from the Sequencer into Native:

```rust
pub struct LibfuncProfileSummary {
    pub libfunc_idx: ConcreteLibfuncId,
    pub samples: u64,
    pub total_time: u64,
    pub average_time: f64,
    pub std_deviation: f64,
    pub quartiles: [u64; 5],
}
```

And have something like the following function for processing.

```rust
fn summarize_profile(program: &Program, profile_data: &[StatementIdx, u64]) -> Vec<LibfuncProfileSummary>;
```

We can then call this from Sequencer / cairo-native-run.

programs/examples/brainfuck.cairo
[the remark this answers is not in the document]
FrancoGiachetta: This was done due to deprecations. We import the trait instead of the method is self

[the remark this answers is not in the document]
JulianGCalderon: Oh I see, is it because we updated Cargo lock? I think its not necessary to update the Cargo lock in this PR.

src/metadata/profiler.rs
JulianGCalderon: The main benefit would be to separate concerns. The user of the library can choose how to process the samples, and we just provide a helper function for summarizing the data. We don't need to tie the processing of the samples to the capturing of them.

Also, I doesn't need to be a method. It would be a standalone function.

The other benefit would be to remove `program` from `ProfileImpl`, as its only used for processing.

FrancoGiachetta: I see you point. However being the pass a `Program` as a parameter would mean that you can pass any `Program`, which could lead to incorrect results. Having the Program "attach" ensures this won't happen. Would receiving a closure as a parameter work?

src/libfuncs.rs
```diff
@@ -279,6 +279,13 @@ where
 
     pub branches: Vec<(&'this Block<'ctx>, Vec<BranchArg<'ctx, 'this>>)>,
     pub results: Vec<Vec<Cell<Option<Value<'ctx, 'this>>>>>,
+
+    #[cfg(feature = "with-libfunc-profiling")]
+    pub profiler: Option<(
+        crate::metadata::profiler::ProfilerMeta,
+        cairo_lang_sierra::program::StatementIdx,
+        (Value<'ctx, 'this>, Value<'ctx, 'this>),
+    )>,
```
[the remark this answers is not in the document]
FrancoGiachetta: Done

src/metadata/profiler.rs
```diff
@@ -0,0 +1,431 @@
...
+    collections::{HashMap, HashSet},
+    ffi::c_void,
+    ptr,
+    sync::{LazyLock, Mutex},
+};
+
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub enum ProfilerBinding {
+    PushStmt,
+    TraceId,
```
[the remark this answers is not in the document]
FrancoGiachetta: Done

src/metadata/profiler.rs
```diff
@@ -0,0 +1,431 @@
...
+        let value = block.shli(value_hi, k32, location)?;
+        let value = block.append_op_result(arith::ori(value, value_lo, location))?;
+
+        // Values
+        // 1. Timestamp
+        // 2. CPU's id core in which the program is running
+        //
+        // We use the last value to ensure that both the initial and then end timestamp of
+        // a libfunc's execution where calculated by the same core. This is to avoid gathering
+        // invalid data
```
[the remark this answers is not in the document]
FrancoGiachetta: Done

src/metadata/profiler.rs
```diff
@@ -0,0 +1,431 @@
...
+                .add_results(&[i64_ty])
+                .build()?,
+        )?;
+        let core_idx = block.const_int(context, location, 0, 64)?;
+
+        Ok((value, core_idx))
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    /// Receives two timestamps, if they were originated in the same statement index
```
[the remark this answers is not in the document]
FrancoGiachetta: Done

src/metadata/profiler.rs
JulianGCalderon: What do you mean by a closure?

src/metadata/profiler.rs
```diff
@@ -0,0 +1,431 @@
...
+            ProfilerBinding::PushStmt => ProfileImpl::push_stmt as *const (),
+            ProfilerBinding::TraceId => ptr::null(),
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct ProfilerMeta {
+    active_map: RefCell<HashSet<ProfilerBinding>>,
+    _private: (),
```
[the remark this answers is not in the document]
FrancoGiachetta: I think it was simply to avoid constructing the struct directly. However I think it is useless now since is has another private field. I'll change it

src/metadata/profiler.rs
JulianGCalderon: I think this design debate is subjective. I don't want to block on this. We can leave it as is, or seek an external opinion. 

Regardless. I don't see any benefit for defining a closure for processing the raw data, instead of exposing the raw data to the user directly. Both achieve the same result, but the closure approach adds unnecessary complexity IMO.

I think something like the following is simpler:

```rust
pub struct ProfilerImpl {}

impl ProfilerImpl {
  pub fn get_profile(&self) -> Profile;
}

type Profile = Vec<Sample>;
pub struct Sample {
  pub statement: StatementIdx,
  pub measure: SampleMeasure,
}
pub enum SampleMeasure {
  DeltaNs(u128),
  Unit,
}
```

This way we just provide a way for accessing the whole program profile, allowing us to:
- Calculate statistics for a libfunc (avg, mean, dev).
- Calculate distribution of a libfunc (does it always take X time, or its execution time varies a lot).
- As we don't aggregate results prematurely, we also have information on the order that this samples were recorded. Note that if we group by libfunc we lose this information.

If its something we do very often, we can also provide a helper for calculating the summary:

```rust
pub struct LibfuncProfileSummary {
    pub libfunc_idx: ConcreteLibfuncId,
    pub samples: u64,
    pub total_time: u64,
    pub average_time: f64,
    pub std_deviation: f64,
    pub quartiles: [u64; 5],
}

impl Profile {
    pub fn summarize(&self, program: &Program) -> Vec<LibfuncProfileSummary>;
}
```

> However being able to pass a Program as a parameter would mean that you can pass any Program, which could lead to incorrect results.

True, but the current solution does not tackle this. The type `ConcreteLibfuncId` depends on the program registry. For the user to correctly inspect that libfunc, it needs the registry of the program. Also, the user needs to create `ProfilerImpl` at the start of the execution, an can also use the wrong `Program`.

gabrielbosio: If processing is something it is done at the end of the `summarize_profiles` we can separate that function in two parts. That way we remove the need of using a closure. The flow would look like this:

1. The user calls `summarize_profiles`. It returns a value.
2. The user does something with that returned value.

I think it is simpler than:

1. The user implements a closure that receives a value that will be processed by `summarize_profiles` in some way.
2. The user calls `summarize_profiles` passing that closure as an argument. It returns a value dependent of the user-defined closure.

Maybe it is because I like to think data flow more linearly. Maybe a closure would be necessary if the processing is being done in the middle of the `summarize_profiles` function (Being the `processor` a hook in that case).

gabrielbosio: It is true that the complexity is being moved to the creation of other structs. Still I prefer to think this more linearly in this case.

src/metadata/profiler.rs
```diff
@@ -0,0 +1,395 @@
...
+//! Once the program execution finished and the information was gathered, a `summarize_profiles` can be called. It
+//! exepects a closure to process this information.
+//!
+//! As well as with the trace-dump fature, in the context of starknet contracts, we need to add support for building
+//! profiles for multiple programs. To do so, we need two important elements, which must be set before every contract
+//! execution:
+//!
+//! 1. A glbal static hashmap to map every profile id to its respective profile summary. See `LIBFUNC_PROFILE`.
+//!
+//! 2. A counter to track the ID of the current libfunc profile, which gets updated every time we switch to another
```
JulianGCalderon, line 20:
```suggestion
//! 2. A counter to track the ID of the current profiler, which gets updated every time we switch to another
```

src/metadata/profiler.rs
```diff
@@ -0,0 +1,395 @@
...
+    },
+    ir::{
+        attribute::{FlatSymbolRefAttribute, StringAttribute, TypeAttribute},
+        operation::OperationBuilder,
+        r#type::{IntegerType, MemRefType},
+        Attribute, Block, BlockLike, Identifier, Location, Module, Region, Value,
+    },
+    Context,
+};
+#[cfg(feature = "with-libfunc-profiling")]
```
JulianGCalderon, line 47: This feature gate should not be required, as this file is gated entirely.

src/metadata/profiler.rs
```diff
@@ -0,0 +1,395 @@
...
+                MemRefType::new(IntegerType::new(context, 64).into(), &[], None, None),
+                location,
+            ))
+            .unwrap();
+
+        block.append_op_result(memref::load(trace_profile_ptr, &[], location))
+    }
+
+    /// Get the timestamp
+    /// Values
```
JulianGCalderon, line 172:
```suggestion
    /// Gets the current timestamp.
    ///
    /// The values returned are:
```

src/metadata/profiler.rs
```diff
@@ -0,0 +1,395 @@
...
+        let value_hi = block.extui(value_hi, i64_ty, location)?;
+        let value_lo = block.extui(value_lo, i64_ty, location)?;
+        let value = block.shli(value_hi, k32, location)?;
+        let value = block.append_op_result(arith::ori(value, value_lo, location))?;
+
+        Ok((value, core_idx))
+    }
+
+    /// Get the timestamp
+    /// Values
```
JulianGCalderon, line 228: /// Gets the current timestamp.
    ///
    /// The values returned are:

src/metadata/profiler.rs
```diff
@@ -0,0 +1,395 @@
...
+    /// 1. Timestamp
+    /// 2. CPU's id core in which the program is running (only for x86 arch)
+    ///
```
JulianGCalderon, line 231: Please add a comment in the function signature, explaining that the core idx cannot be retreived in aarch64, and therefore we always return the same value: `0`.

src/metadata/profiler.rs
```diff
@@ -0,0 +1,395 @@
...
+        Ok(())
+    }
+}
+
+/// Represents a libfunc's profile. A libfunc profile maps libfuncs id
+/// to a tuple with a vector of deltas (*1) and extra counts (*2).
+/// *1 Each delta refers to the execution time of that libfunc call
+/// *2 `extra_count` is a count of libfunc calls whose execution time
+/// couldn't be calculated.
+pub type LibfuncProfiles = HashMap<ConcreteLibfuncId, (Vec<u64>, u64)>;
```
JulianGCalderon, line 327: Please change the tuple for a structure:

```rust
/// Represents the entire profile of the execution.
///
///It maps the libfunc ID to a libfunc profile.
type Profile = HashMap<ConcreteLibfuncId, LibfuncProfileData)>;

/// Represents the profile data for a particular libfunc.
struct LibfuncProfileData {
    /// A vector of execution times, for each time the libfunc was executed.
    deltas: Vec<u64>,
    /// If the time delta for a particular execution could not be gathered,
    /// we just increase `extra_counts` by 1.
    extra_counts: u64,
}
```

JulianGCalderon: Also, document what is the unit of the deltas vector. Is it in nanoseconds?

src/metadata/profiler.rs
```diff
@@ -0,0 +1,395 @@
...
+
+pub static LIBFUNC_PROFILE: LazyLock<Mutex<HashMap<u64, ProfileImpl>>> =
+    LazyLock::new(|| Mutex::new(HashMap::new()));
+
+#[derive(Default)]
+pub struct ProfileImpl {
```
JulianGCalderon, line 333:
```suggestion
pub struct ProfilerImpl {
```

src/metadata/profiler.rs
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    /// Process profiling results.
+    ///
+    /// Receives a closure with the flowing paramaters `(libfunc_id, (vec<delta_time>, extra_count))` to
+    /// process profiles information.
+    ///
+    /// `extra_count`: counter of libfunc calls whose execution time
+    /// hasn't been calculated for being invalid.
+    ///
```
JulianGCalderon, line 363: This comments are outdated. I think they could be removed. If we document de data type correctly, there is no need to repeat the meaning of the fields here.

src/metadata/profiler.rs
```diff
@@ -0,0 +1,395 @@
...
+/// *2 `extra_count` is a count of libfunc calls whose execution time
+/// couldn't be calculated.
+pub type LibfuncProfiles = HashMap<ConcreteLibfuncId, (Vec<u64>, u64)>;
+
+pub static LIBFUNC_PROFILE: LazyLock<Mutex<HashMap<u64, ProfileImpl>>> =
+    LazyLock::new(|| Mutex::new(HashMap::new()));
+
+#[derive(Default)]
+pub struct ProfileImpl {
+    pub trace: Vec<(StatementIdx, u64)>,
```
JulianGCalderon, line 334:
```suggestion
    /// The samples recorded by the profiler. A value of `u64::MAX` implies
    /// that the delta time for a statement could not be gathered.
    pub samples: Vec<(StatementIdx, u64)>,
```

src/metadata/profiler.rs
```diff
@@ -0,0 +1,395 @@
...
+            eprintln!("Could not find libfunc profiler!");
+            return;
+        };
+
+        profiler
+            .trace
+            .push((StatementIdx(statement_idx as usize), tick_delta));
+    }
+
+    /// Process profiling results.
```
JulianGCalderon, line 356: I'd rename this comment to: "Returns the execution profile, grouped by libfunc".

src/metadata/profiler.rs
```diff
@@ -0,0 +1,395 @@
...
+    ///
+    /// Receives a closure with the flowing paramaters `(libfunc_id, (vec<delta_time>, extra_count))` to
+    /// process profiles information.
+    ///
+    /// `extra_count`: counter of libfunc calls whose execution time
+    /// hasn't been calculated for being invalid.
+    ///
+    pub fn get_profiles(&self, sierra_program: &Program) -> LibfuncProfiles {
```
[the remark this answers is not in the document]
FrancoGiachetta: Mmm, I saw a profile a profile as an individual Libfunc. But I'll change it if you prefer

src/bin/cairo-native-run.rs
```diff
@@ -188,6 +221,53 @@ fn main() -> anyhow::Result<()> {
...
+
+        assert_eq!(profile.values().len(), 1);
+
+        let profile = profile.values().next().unwrap();
+
+        if let Some(profiler_output_path) = args.profiler_output {
+            let mut output = File::create(profiler_output_path)?;
+
+            let raw_profiles = profile.get_profile(&sierra_program);
+            let mut processed_profiles = process_profiles(raw_profiles);
```
JulianGCalderon, line 240: To keep consistency, we should use the singular here `raw_profile`, `processed_profile`, etc.

src/metadata/profiler.rs
```diff
@@ -0,0 +1,415 @@
...
+//!    we executed the given statement.
+//!
+//! Once the program execution finished and the information was gathered, the `get_profile` method can be called.
+//! It groups the samples by libfunc, and returns all data related to each libfunc.
+//!
+//! As well as with the trace-dump feature, in the context of Starknet contracts, we need to add support for building
+//! profiles for multiple executions. To do so, we need two important elements, which must be set before every contract
+//! execution:
+//!
+//! 1. A golbal static hashmap to map every profile ID to its respective profiler. See `LIBFUNC_PROFILE`.
```
JulianGCalderon, line 19:
```suggestion
//! 1. A global static hashmap to map every profile ID to its respective profiler. See `LIBFUNC_PROFILE`.
```

src/metadata/profiler.rs
```diff
@@ -0,0 +1,415 @@
...
+//!
+//! 2. A counter to track the ID of the current profiler, which gets updated every time we switch to another
+//!    contract. Since a contract can call other contract's, we need a way of restoring the counter after every execution.
```
JulianGCalderon, line 22:
```suggestion
//!    contract. Since a contract can call other contracts, we need a way of restoring the counter after every execution.
```

src/metadata/profiler.rs
```diff
@@ -0,0 +1,415 @@
...
+
+        block.append_op_result(memref::load(trace_profile_ptr, &[], location))
+    }
+
+    /// Gets the current timestamp.
+    ///
+    /// The values returned are:
+    /// 1. Timestamp: CPU cycles since its reset.
+    /// 2. CPU's id core in which the execution is running (only for x86 arch).
+    ///    In case of arm, 0 is always return as there's no way to know in which
```
JulianGCalderon, line 177:
```suggestion
    ///    In case of arm, 0 is always returned as there's no way to know in which
```

src/metadata/profiler.rs
```diff
@@ -0,0 +1,415 @@
...
+
+        Ok((value, core_idx))
+    }
+
+    /// Gets the current timestamp.
+    ///
+    /// The values returned are:
+    /// 1. Timestamp: CPU cycles since its reset.
+    /// 2. CPU's id core in which the program is running (only for x86 arch).
+    ///    In case of arm, 0 is always return as there's no way to know in which
```
JulianGCalderon, line 236:
```suggestion
    ///    In case of arm, 0 is always returned as there's no way to know in which
```

src/metadata/profiler.rs
```diff
@@ -0,0 +1,415 @@
...
+
+        Ok(())
+    }
+}
+
+/// Represents a libfunc's profile. A libfunc profile maps libfuncs id
+/// to a tuple with a vector of deltas (*1) and extra counts (*2).
+/// *1 Each delta refers to the execution time of that libfunc call
+/// *2 `extra_count` is a count of libfunc calls whose execution time
+/// couldn't be calculated.
```
JulianGCalderon, line 333: This comment is outdated.

src/metadata/profiler.rs
```diff
@@ -0,0 +1,415 @@
...
+/// Represents the entire profile of the execution.
+///
+///It maps the libfunc ID to a libfunc profile.
```
JulianGCalderon, line 336:
```suggestion
/// It maps the libfunc ID to a libfunc profile.
```

JulianGCalderon (changes requested):

src/metadata/profiler.rs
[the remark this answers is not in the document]
FrancoGiachetta: It is usable but there's a possibility that there could be noise in the results, which is not the case in x86.

src/bin/cairo-native-run.rs
```diff
@@ -205,3 +298,92 @@ fn main() -> anyhow::Result<()> {
...
 }
+
+#[cfg(feature = "with-libfunc-profiling")]
+struct LibfuncProfileSummary {
+    pub libfunc_idx: ConcreteLibfuncId,
+    pub samples: u64,
+    pub total_time: Option<u64>,
+    pub average_time: Option<f64>,
+    pub std_deviation: Option<f64>,
+    pub quartiles: Option<[u64; 5]>,
```
JulianGCalderon, line 309: In which situation do we return `None`?

src/bin/cairo-native-run.rs
```diff
@@ -188,6 +221,66 @@ fn main() -> anyhow::Result<()> {
...
+
+        assert_eq!(profile.values().len(), 1);
+
+        let profile = profile.values().next().unwrap();
+
+        if let Some(profiler_output_path) = args.profiler_output {
+            let mut output = File::create(profiler_output_path)?;
+
+            let raw_profile = profile.get_profile(&sierra_program);
+            let mut processed_profile = process_profiles(raw_profile);
```
JulianGCalderon, line 240:
```suggestion
            let mut processed_profile = process_profile(raw_profile);
```

src/bin/cairo-native-run.rs
```diff
@@ -188,6 +221,66 @@ fn main() -> anyhow::Result<()> {
...
+                    (total_time, average_time, std_deviation, quartiles)
+                else {
+                    writeln!(output, "    Total Execution Time:   none")?;
+                    writeln!(output, "    Average Execution Time: none")?;
+                    writeln!(output, "    Standard Deviation:     none")?;
+                    writeln!(output, "    Quartiles:              none")?;
+                    writeln!(output)?;
+
+                    continue;
+                };
```
JulianGCalderon, line 273: If only a single field is `None`, this will write `None` in all of them.

src/bin/cairo-native-run.rs
```diff
@@ -205,3 +298,92 @@ fn main() -> anyhow::Result<()> {
...
+                // Compute the quartiles.
+                let quartiles = [
+                    *deltas.first().unwrap(),
+                    deltas[deltas.len() / 4],
+                    deltas[deltas.len() / 2],
+                    deltas[3 * deltas.len() / 4],
+                    *deltas.last().unwrap(),
+                ];
+
+                // Compuite the average.
```
JulianGCalderon, line 361:
```suggestion
                // Compute the average.
```

src/metadata/profiler.rs
```diff
@@ -0,0 +1,411 @@
+#![cfg(feature = "with-libfunc-profiling")]
+//! The libfunc profiling feature is used to generate information about every libfunc executed in a sierra program.
+//!
+//! When this feature is used, the compiler will call the important methods:
+//!
+//!  1. `measure_timestamp`: called before every libfunc execution.
```
JulianGCalderon, line 6:
```suggestion
//! 1. `measure_timestamp`: called before every libfunc execution.
```

src/bin/cairo-native-run.rs
FrancoGiachetta: In x86, if a libfunc is call once and the timestamps calculated are invalid, then `deltas` will be an empty array.

src/bin/cairo-native-run.rs
JulianGCalderon: Wouldn't it be better to check each of them individually? Or it doesn't matter?

FrancoGiachetta: I prefer doing it in one line. I don't want it to panic if no deltas were calculated, so unwrapping is not an option. Checking each value individually would mean I would need to tackle the `None` case for each of them. I think it's simpler and cleaner to leave it in one line.

src/bin/cairo-native-run.rs
FrancoGiachetta: You are, I mess up and used `Some(..)` instead. Fixed it

src/bin/cairo-native-run.rs
JulianGCalderon: If its impossible for one value to be None, but for the other values to be Some, then  We **should** check each value individually. If its not possible for this to happen (either all are None, or all are Some) then we could leave it this way.

JulianGCalderon (approved):
LGTM

src/bin/cairo-native-run.rs
FrancoGiachetta: All this values depend on the fact that `deltas` is not empty. So it is not.

PR #965, opened 2024-12-13
Improve array documentation (and minor bug)

JulianGCalderon:
- Adds comments to make the code easier to understand.
- Renames a few variables with clearer names.
- Fixes a minor bug. We were not calculating the array allocation pointer correctly. It was fixed by properly naming variables.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/array.rs
+++ b/src/libfuncs/array.rs
@@ -349,7 +349,7 @@ pub fn build_tuple_from_span<'ctx, 'this>(
             valid_block.const_int(context, location, elem_layout.pad_to_align().size(), 64)?,
             location,
         ))?;
-        let array_ptr = valid_block.gep(
+        let array_data_start_ptr = valid_block.gep(
             context,
             location,
             array_ptr,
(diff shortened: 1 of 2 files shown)


src/libfuncs/array.rs
```diff
@@ -335,28 +361,33 @@ pub fn build_tuple_from_span<'ctx, 'this>(
...
+                    _ => block.memcpy(context, location, array_data_start_ptr, value, value_size),
                 }
 
-                // The following unwrap should be unreachable because an array always has a drop
-                // implementation.
+                // drop the original array (decreasing its reference counter)
                 metadata
                     .get::<DropOverridesMeta>()
-                    .unwrap()
+                    .expect("array always has a drop implementation")
```
FrancoGiachetta, line 390: Why not using ´?´ instead? Just to remove the panic.

FrancoGiachetta (approved):

src/libfuncs/array.rs
azteca1998: This `unwrap` is unreachable. An array always has a drop implementation because we need to free the memory.
Does the `?` operator work directly? Or do we need to `.map_err`?

JulianGCalderon: We would need to use native_assert. I'll change it to avoid panics.

JulianGCalderon: Done

edg-l (approved):

PR #973, opened 2024-12-16
reduce daily block range size to 10

edg-l:
To reduce the timeout chance, daily block running is more a health check, so if it fails due to timeout its pointless

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


FrancoGiachetta (approved):

JulianGCalderon (commented):
Right now we are running 40 workflows, with 25 blocks each, summing up to a total of 1000 blocks each day. If we reduce the range size, we should also increase the number of workflows so that we run more or less the same amount of blocks.

Also, maybe a range size of 10 to little, maybe we could try with 20 and see if it finishes in time.

FrancoGiachetta:
I would say is preferable to have more blocks and leave it to 10 to be sure it won't fail. Looking at the number of transactions that fail, I think 10 blocks range should have us covered. What do you think?

JulianGCalderon:
@FrancoGiachetta 10 blocks will probably work, but if it fails with 20 (or 15), we could make another PR lowering it a bit more, right? There is a limit to the amount of jobs that can be run in a workflow, so there is a benefit on trying to find the maximum range size that will run all blocks, just in case later on we want to increase the size of the daily run.

FrancoGiachetta:
If I'm not mistaken, the limit of workers is 256. But yeah, we could definitely give it a try.

PR #976, opened 2024-12-16
Make daily blocks run sequentially

FrancoGiachetta:
This PR divides the list of blocks in 8 lists of 5 blocks each. It creates 8 workers, each of which depends on the previous to begin execution. It will hopefully avoid both the github and the juno's timeouts. 

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


JulianGCalderon:
@FrancoGiachetta
Although I haven't tested it, we may be able to use the `max-parallel` option, instead of doing it manually:
- https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymax-parallel

FrancoGiachetta:
Thanks! Didn't know of that one, I'll try it.

FrancoGiachetta:
Done

.github/workflows/daily.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       RPC_ENDPOINT_TESTNET: ${{ secrets.RPC_ENDPOINT_TESTNET }}
       RPC_ENDPOINT_MAINNET: ${{ secrets.RPC_ENDPOINT_MAINNET }}
     strategy:
+      max-parallel: 1
```
JulianGCalderon, line 21: We should discuss which number to pick, If we run everything sequentially, the whole workflow would take a long time (maybe more than a day). I'd increase max-parallel to something like 10.

JulianGCalderon (changes requested):

.github/workflows/daily.yml
FrancoGiachetta: You are right, silly me. I'll leave it to 10 to discuss it.

JulianGCalderon (approved):

edg-l (approved):

PR #981, opened 2024-12-19
Add version check for cached AOT contracts.

azteca1998:
For now it'll panic with an assert. It is the user's responsibility to clear the contract cache when upgrading.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/executor/contract.rs
+++ b/src/executor/contract.rs
@@ -119,20 +119,7 @@ pub enum BuiltinType {
 
 impl BuiltinType {
     pub const fn size_in_bytes(&self) -> usize {
-        match self {
-            BuiltinType::Bitwise => 8,
-            BuiltinType::EcOp => 8,
-            BuiltinType::RangeCheck => 8,
-            BuiltinType::SegmentArena => 8,
-            BuiltinType::Poseidon => 8,
-            BuiltinType::Pedersen => 8,
-            BuiltinType::RangeCheck96 => 8,
-            BuiltinType::CircuitAdd => 8,
-            BuiltinType::CircuitMul => 8,
-            BuiltinType::Gas => 16,
-            BuiltinType::System => 8,
-            BuiltinType::BuiltinCosts => 8,
-        }
+        size_of::<u64>()
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/executor/contract.rs
```diff
@@ -119,20 +119,7 @@ pub enum BuiltinType {
...
-            BuiltinType::Bitwise => 8,
-            BuiltinType::EcOp => 8,
-            BuiltinType::RangeCheck => 8,
-            BuiltinType::SegmentArena => 8,
-            BuiltinType::Poseidon => 8,
-            BuiltinType::Pedersen => 8,
-            BuiltinType::RangeCheck96 => 8,
-            BuiltinType::CircuitAdd => 8,
-            BuiltinType::CircuitMul => 8,
-            BuiltinType::Gas => 16,
```
JulianGCalderon, line 132: BuiltinType::Gas has a size of 16. Is this intended?

src/executor/contract.rs
```diff
@@ -279,8 +266,25 @@ impl AotContractExecutor {
...
+
+        let library = Arc::new(unsafe { Library::new(library_path)? });
+        unsafe {
+            let get_version = library
+                .get::<extern "C" fn(*mut u8, usize) -> usize>(b"cairo_native__get_version")?;
+
+            let mut version_buffer = [0u8; 16];
+            let version_len = get_version(version_buffer.as_mut_ptr(), version_buffer.len());
+
+            let target_version = env!("CARGO_PKG_VERSION");
```
JulianGCalderon, line 278: This would fail if we don't change cargo pkg version, right? Is it possible to use the git rev?

src/executor/contract.rs
azteca1998: This was correct when gas was still an `u128`, but we changed that a while ago. It should be correct now.

src/executor/contract.rs
azteca1998: Cargo has no notion of git revisions. It'll work between different versions only.

JulianGCalderon: Would executing `git rev-parse HEAD` on build time work?

azteca1998: Not if we decide to not use git. Since cargo and git are not related (they are independent and know nothing of each other) I don't think it's a good idea to mix them.

Also we'll be depending on the shell environment being properly configured and git installed. You can download the compressed source code without git stuff from the releases page so you don't actually need git for anything; this would break the build for systems that build without git installed.

edg-l (approved):

JulianGCalderon (approved):

PR #985, opened 2024-12-23
Set Runtime as artifact dependency

gabrielbosio:
To build Native run `cargo build -Z bindeps`.

This only works with Rust nightly.


FrancoGiachetta:
Little more info about the PR:
1. It Replaces the way we link the runtime. Now there's no need to have an env var point wherever the runtime library was build.
2. It adds a `.cargo/config.toml` to make `bindeps` default
3. Sets the rust-toolchain to nightly in order to be able to use bindeps.

gabrielbosio:
#1051 already solves this.

PR #986, opened 2024-12-26
update cairo to 2.10rc-0

FrancoGiachetta:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


azteca1998 (dismissed):

src/libfuncs/bounded_int.rs
```diff
@@ -699,6 +702,41 @@ fn build_constrain<'ctx, 'this>(
...
+/// Generate MLIR operations for the `bounded_int_trim` libfunc.
+fn build_trim<'ctx, 'this>(
+    context: &'ctx Context,
+    _registry: &ProgramRegistry<CoreType, CoreLibfunc>,
+    entry: &'this Block<'ctx>,
+    location: Location<'ctx>,
+    helper: &LibfuncHelper<'ctx, 'this>,
+    _metadata: &mut MetadataStorage,
+    info: &BoundedIntTrimConcreteLibfunc,
+) -> Result<()> {
```
JulianGCalderon, line 714: Could you add a comment describing what the libfunc does?

JulianGCalderon: You could use the array documentation as an example: https://github.com/lambdaclass/cairo_native/blob/e41f7b11c513f23c600898011e943217eeaf509a/src/libfuncs/array.rs#L154-L162

src/libfuncs/gas.rs
```diff
@@ -153,6 +155,78 @@ pub fn build_withdraw_gas<'ctx, 'this>(
...
 
+pub fn build_redeposit_gas<'ctx, 'this>(
+    context: &'ctx Context,
+    _registry: &ProgramRegistry<CoreType, CoreLibfunc>,
+    entry: &'this Block<'ctx>,
+    location: Location<'ctx>,
+    helper: &LibfuncHelper<'ctx, 'this>,
+    metadata: &mut MetadataStorage,
+    _info: &SignatureOnlyConcreteLibfunc,
+) -> Result<()> {
```
JulianGCalderon, line 166: Like with the trim libfunc, could you add a comment expaining what it does?

FrancoGiachetta: Done

src/libfuncs/bounded_int.rs
FrancoGiachetta: Done

JulianGCalderon (approved):

JulianGCalderon (dismissed):

edg-l (dismissed):

edg-l (dismissed):

src/types.rs
```diff
@@ -122,6 +122,14 @@ pub trait TypeBuilder {
         &self,
         registry: &ProgramRegistry<CoreType, CoreLibfunc>,
     ) -> Result<bool, Self::Error>;
+
+    /// Return whether the type is signed, either directly or indirectly (ex. through
+    /// `Const`).
+    fn is_signed(
+        &self,
+        registry: &ProgramRegistry<CoreType, CoreLibfunc>,
+    ) -> Result<bool, Self::Error>;
```
azteca1998, line 131: Instead of this, we should check whether `ty.integer_range()` can have negative values. This can be done by checking the lower bound's sign. If negative, then the integer type is signed.

FrancoGiachetta: I see, I'll do that then. Thanks!

FrancoGiachetta: Sorry, I forgot about something. I'm not checking if it is negative, in fact it does not matter wether it is or not. If its type signed, then we should handle it that way. That's why a created that method

FrancoGiachetta: Done

azteca1998 (approved):

JulianGCalderon (approved):

PR #1029, opened 2025-01-07
fix runtime version check

FrancoGiachetta:
When checking the runtime's version, it occurs that version's buffer does not have the same size as the target buffer. Due to this, the function `copy_from_slice` fails when trying to copy the contents of the version's buffer into the target's one

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


runtime/src/lib.rs
```diff
@@ -42,10 +42,12 @@ pub unsafe extern "C" fn cairo_native__get_version(target: *mut u8, length: usiz
     let target = slice::from_raw_parts_mut(target, length);
 
     let version = env!("CARGO_PKG_VERSION");
-    assert!(length > version.len(), "version buffer not big enough");
+    assert!(
+        length == version.len(),
+        "target buffer should have the same size as version buffer"
+    );
```
JulianGCalderon, line 48: I'm not sure this is the right approach. We would need to know beforehand the length of the version string, which will vary between versions.

Maybe there is an alternative method to `copy_from_slice` that doesn't require both slices to be of equal length? Or a way to resize the slice to match the string length.

We should copy the contents of the version string into the received buffer, inserting a \0 at the end.

JulianGCalderon (changes requested):

runtime/src/lib.rs
FrancoGiachetta: We could first check if `length > version.len()` and then use `version`'s lenght when defining `target`.

FrancoGiachetta: I've tried that approach, I think it's much better.

azteca1998:
The `'\0'` is there because C strings work that way. Please do not remove it.

The condition `length > version.len()` should ensure that `target` has enough space for the version, including the final `\0` byte.

The only thing that needs to be changed is the `length` by `version.len()`, which you already did. Everything else is not necessary. You can move the `slice::from_raw_parts` until after the check, but it shouldn't affect anything because we're not accessing it until we've checked we can do so.

FrancoGiachetta:
Oh, you are right. I removed it in the previous approach, which match the same length as `version` and so inserting '\0' would overwrite the last byte. I'll change it.

edg-l (approved):

azteca1998 (approved):

JulianGCalderon (approved):

PR #1051, opened 2025-01-14
Implement runtime with dynamic dispatch

JulianGCalderon:
Implements runtime with dynamic dispatch.

Depends on #1052

Previously, runtime function were defined in a different crate `cairo_native_runtime`, compiled as a static library (`.a`), and linked to the starknet contract.

This PR gets rid of the runtime static library completely, by defining a global for each runtime function. On execution, it will access the global to find the runtime function. This global is originally empty, but when loading the executor, the caller sets the globals, making them point to the correct functions (located in the callers binary).

This PR is a POC and should be tested thoroughly. As this new approach adds an additional level of indirection when calling the runtime, we should check that it doesn't introduce benchmark regressions.

## Implementation Details

The `build_function` method declares the global if not yet declared, accesses it, and calls the function.

The `setup_runtime` function sets the value of all defined globals with their corresponding location. It is called when loading the executors.

Since we are using rust's runtime, instead of contract's runtime, we can call `set_costs_builtin` directly, instead of looking for the symbol.

I removed the hyperfine standalone AOT benches, as adapting it to the new dynamic runtime is difficult, and probably not worth it.

## Roadmap

- [x] Adapt all runtime bindings
- [x] Remove original runtime link (AOT)
- [x] Remove original runtime register (JIT)
- [x] Compare execution with VM on many blocks
  - From 900000 to 900484 without diffs
  - From 1000000 to 1000500 without diffs
- [x] Measure performance: 8% to 11% improvement
  - From 900000 to 900010: 8%
    - With static Runtime: 1.78s
    - With dynamic Runtime: 1.64s
  - From 900010 to 900020: 9%
    - With static Runtime: 2.81s
    - With dynamic Runtime: 2.57s
  - From 900020 to 900030: 8%
    - With static Runtime: 3.89s
    - With dynamic Runtime: 3.59s
  - From 900030 to 900040: 11%
    - With static Runtime: 5.47s
    - With dynamic Runtime: 4.91s
- [x] Measure compiled programs size: 20X improvement
  - Average complied library with static runtime: 13MiB
  - Average compiled library with dynamic runtime: 450KiB

Why is it faster? We should take a closer look at it.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/executor/contract.rs
```diff
@@ -268,27 +272,14 @@ impl AotContractExecutor {
...
-            let target_version = env!("CARGO_PKG_VERSION");
-            assert_eq!(
-                &version_buffer[..version_len],
-                target_version.as_bytes(),
-                "aot-compiled contract version mismatch"
-            );
-        };
-
-        Ok(Self {
+        let executor = Self {
```
azteca1998, line 275: Can't `Self::new` be used here too?

azteca1998 (dismissed):

src/executor/contract.rs
JulianGCalderon: `Self::new` is used for compiling a new program, while `Self::load` is used for loading an already compiled library. In this case we can't use `Self::new`.

azteca1998: I see, I missed all the other stuff that `new` does (it was collapsed).

edg-l (approved):

gabrielbosio (approved):

edg-l (approved):

azteca1998 (approved):

PR #959, opened 2024-12-05
Add `SafeRunner` to catch `SIGSEGV` and grow the stack automagically.

azteca1998:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
- [ ] Integration tests added.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.

PR #1053, opened 2025-01-15
Add segfault catcher.

azteca1998:
Simplified version of #959. Does not attempt to grow the stack, just catches segfaults and returns an error.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


edg-l (approved):
Is it possible to add a test that segfaults? Or too tricky?

azteca1998:
It is **technically** possible, but it'd require a way to initialize the signal handler globally when testing, potentially messing up the test harness's signal handler (ex. when using nextest).

Instead, if we just initialized every test then we'd end up initializing it multiple times, which could also cause problems.

src/executor/contract.rs
```diff
@@ -436,14 +436,15 @@ impl AotContractExecutor {
...
+        crate::utils::safe_runner::run_safely(|| unsafe {
             invoke_trampoline(
                 function_ptr,
                 invoke_data.as_ptr().cast(),
                 invoke_data.len() >> 3,
                 ret_registers.as_mut_ptr(),
             );
-        }
+        })
+        .map_err(Error::SafeRunner)?;
```
JulianGCalderon, line 447: Should we gate this under a feature flag? Or maybe provide an API instead of using directly?

src/utils/safe_runner.rs
```diff
@@ -0,0 +1,118 @@
...
+    };
+
+    STATE.with(|x| unsafe { ptr::write(x.get(), prev_state) });
+    result
+}
+
+unsafe extern "C" fn segfault_handler(_sig: c_int, _info: &siginfo_t, _context: &mut ucontext_t) {
+    match STATE.with(|x| &mut *x.get()) {
+        SafeRunnerState::Inactive => libc::abort(),
+        SafeRunnerState::Active(jmp_buf) => longjmp(jmp_buf.as_mut_ptr().cast(), 1),
```
JulianGCalderon, line 116: The `1` value is then returned as `jmp_ret` in the `setjmp` call, right? If so, could you add a small comment mentioning it?

JulianGCalderon, line 116: After the long jump, the stack is unwinded, which means that stack memory is not leaked, right?

src/utils/safe_runner.rs
```diff
@@ -0,0 +1,118 @@
...
+        );
+        assert_eq!(
+            sigaltstack(
+                &stack_t {
+                    ss_sp: STACK.with(|x| x.get()).cast(),
+                    ss_flags: 0,
+                    ss_size: SIGSTKSZ,
+                },
+                null_mut(),
+            ),
```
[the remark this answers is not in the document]
azteca1998: Yes, if the stack pointer is currently overflowed, using it from the signal handler would cause a double segfault.

src/utils/safe_runner.rs
azteca1998: Yes, it does exactly that. Even if you pass a 0 to `longjmp` it'll return a 1. In other words, the only time it'll return 0 is when `setjmp` is initializing stuff, but never when returning from a `longjmp`.

edg-l (dismissed):

src/utils/safe_runner.rs
```diff
@@ -0,0 +1,139 @@
+//! # Safe runner
+//!
+//! The safe runner provides a way to run contracts without any risk of crashing due to invalid
+//! memory accesses, including stack overflows. The same mechanism can also be used to abort the
+//! current Cairo program execution, although it will probably leak some memory.
+//!
+//! It works by setting a signal handler for the `SIGSEGV` signal, which is generated by the
+//! operating system on invalid memory accesses. This signal handler is global, therefore it must be
+//! set by the application, rather than the library, by invoking `setup_safe_runner`.
```
JulianGCalderon, line 9: If the executor is called without calling `setup_safe_runner` it will fail, right? 

In that case, I think we should document this in the executor API also, as here its easy to miss.

azteca1998: If the safe runner is used without `setup_safe_runner` it should still run the code but it won't catch any segfault.

JulianGCalderon (approved):

edg-l (approved):

PR #1055, opened 2025-01-16
Remove missing todos

edg-l:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -758,7 +758,7 @@ fn compile_func(
                                             }
                                         })
                                         .collect::<Result<Vec<_>, _>>()?,
-                                    None => native_panic!("should be Some always"),
+                                    None => native_panic!("not yet implemented"),
                                 };
 
                                 block.append_operation(cf::cond_br(
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -758,7 +758,7 @@ fn compile_func(
                                             }
                                         })
                                         .collect::<Result<Vec<_>, _>>()?,
-                                    None => todo!(),
+                                    None => native_panic!("should be Some always"),
                                 };
 
                                 block.append_operation(cf::cond_br(
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -635,9 +635,9 @@ fn parse_result(
         | CoreTypeConcrete::StarkNet(_)
         | CoreTypeConcrete::Uint128MulGuarantee(_)
         | CoreTypeConcrete::Circuit(_)
-        | CoreTypeConcrete::RangeCheck96(_) => todo!(),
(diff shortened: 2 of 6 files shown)


JulianGCalderon (approved):

src/compiler.rs
```diff
@@ -758,7 +758,7 @@ fn compile_func(
                                             }
                                         })
                                         .collect::<Result<Vec<_>, _>>()?,
-                                    None => todo!(),
+                                    None => native_panic!("should be Some always"),
```
azteca1998, line 761: I'm not sure this should always be `Some(_)`, otherwise it'd be an `unreachable!()`, not a `todo!()`. Maybe this error message is more descriptive:
```suggestion
                                    None => native_panic!("not yet implemented"),
```
Yes, it's basically a `todo!()` but without the panic.

src/executor.rs
```diff
@@ -635,9 +635,9 @@ fn parse_result(
         | CoreTypeConcrete::StarkNet(_)
         | CoreTypeConcrete::Uint128MulGuarantee(_)
         | CoreTypeConcrete::Circuit(_)
-        | CoreTypeConcrete::RangeCheck96(_) => todo!(),
+        | CoreTypeConcrete::RangeCheck96(_) => native_panic!("not yet implemented as results"),
```
azteca1998, line 638: The `RangeCheck96` is a builtin, therefore should go in the builtins category (the block just above, after `RangeCheck`).

src/libfuncs/int.rs
```diff
@@ -636,7 +637,7 @@ fn build_square_root<'ctx, 'this>(
             CoreTypeConcrete::Uint32(_) => (32, 16),
             CoreTypeConcrete::Uint64(_) => (64, 32),
             CoreTypeConcrete::Uint128(_) => (128, 64),
-            _ => unreachable!(),
+            _ => native_panic!("unhandled type in int square root"),
```
azteca1998, line 640: Rather than unhandled it's invalid because there are no more integers.
An example: it doesn't make sense to compute the integer square root of a struct.
```suggestion
            _ => native_panic!("invalid value type in int square root"),
```

src/types/circuit.rs
```diff
@@ -280,7 +281,7 @@ pub fn layout(
             Ok(layout)
         }
         CircuitTypeConcrete::CircuitPartialOutputs(_) => {
-            todo!("CircuitPartialOutputs is noop for now")
+            native_panic!("CircuitPartialOutputs is noop for now");
```
azteca1998, line 284: This type does have a layout, even though it's a placeholder.
```suggestion
            Ok(Layout::new::<()>())
```

JulianGCalderon (approved):

gabrielbosio (approved):

PR #1057, opened 2025-01-16
Change tests to use sierra generator

FrancoGiachetta:
This PR uses the sierra generator to test libfuncs. This is to prevent us from any possible change in the sierra generator which could generate false positives in tests. Some tests were left untouched either because the libfunc they were targeting was already being tested in another test, or because they were using the libfunc explicitly.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/libfuncs/bytes31.rs
```diff
@@ -162,39 +162,53 @@ pub fn build_from_felt252<'ctx, 'this>(
...
+        else {
+            panic!("Invalid result type");
+        };
+        let result = run_sierra_program(&BYTES31_TO_FELT, &[*value]).return_value;
+
+        assert_eq!(Value::Felt252(2.into()), result);
+
+        let result = run_sierra_program(&BYTES31_TRY_FROM_FELT, &[Felt::MAX.into()]).return_value;
+
+        assert_eq!(jit_enum!(1, jit_struct!()), result);
```
JulianGCalderon, line 212: Could you add the TO_FELT call for this case also?

src/libfuncs/felt252_dict.rs
```diff
@@ -228,6 +250,24 @@ mod test {
...
+
+            generator.build(&[GenericArg::Type(u64_ty)])
+        };
+
+        let dict = Value::Felt252Dict {
+            value: HashMap::new(),
+            debug_name: None,
+        };
+
+        run_sierra_program(&program, &[dict]);
```
JulianGCalderon, line 268: What is the result type of the libfunc?

src/libfuncs/uint256.rs
```diff
@@ -1165,19 +1158,34 @@ mod test {
...
                 ],
-                result,
             )
+            .return_value;
+
+            assert_eq!(expected, result);
         }
 
-        let none = jit_enum!(1, jit_struct!());
+        // garantees are represented as Value::Null
```
JulianGCalderon, line 1174:
```suggestion
        // guarantees are represented as Value::Null
```

src/libfuncs/felt252.rs
```diff
@@ -509,63 +503,43 @@ pub mod test {
...
+            x
+        }
 
         // Division by zero is expected to panic.
-        assert_panics(f("0"), f("0"));
-        assert_panics(f("1"), f("0"));
-        assert_panics(f("-2"), f("0"));
+        assert_eq!(r(f("0"), f("0")), f("0"));
+        assert_eq!(r(f("1"), f("0")), f("0"));
+        assert_eq!(r(f("-2"), f("0")), f("0"));
```
JulianGCalderon, line 520: Why division by zero no longer panics?

src/libfuncs/enum.rs
JulianGCalderon, line 1: The programs in this file are very complex, and the sierra generator is not easy to follow. Maybe we should leave them as sierra code?

src/libfuncs/cast.rs
```diff
@@ -390,299 +390,571 @@ pub fn build_upcast<'ctx, 'this>(
...
-                    (downcast(v128),),
-                )
-            }
+        static ref DOWNCAST_U128_U128: Program = {
+            let mut generator = SierraGenerator::<DowncastLibfunc>::default();
+
+            let from_ty = generator.push_type_declaration::<Uint128Type>(&[]).clone();
+            let to_ty = generator.push_type_declaration::<Uint128Type>(&[]).clone();
+
+            generator.build(&[GenericArg::Type(from_ty), GenericArg::Type(to_ty)])
```
JulianGCalderon, line 419: Maybe we could have a generic function that builds this, instead of having a constant for each possibility?

src/values.rs
```diff
@@ -749,9 +759,7 @@ impl Value {
                 CoreTypeConcrete::Uint32(_) => Self::Uint32(*ptr.cast::<u32>().as_ref()),
                 CoreTypeConcrete::Uint64(_) => Self::Uint64(*ptr.cast::<u64>().as_ref()),
                 CoreTypeConcrete::Uint128(_) => Self::Uint128(*ptr.cast::<u128>().as_ref()),
-                CoreTypeConcrete::Uint128MulGuarantee(_) => {
-                    native_panic!("todo: implement uint128mulguarantee from_ptr")
-                }
```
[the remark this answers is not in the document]
FrancoGiachetta: Yes it could be added

src/libfuncs/felt252_dict_entry.rs
```diff
@@ -193,22 +193,6 @@ pub fn build_finalize<'ctx, 'this>(
...
-            fn run_test() -> u32 {
-                let mut dict: Felt252Dict<u32> = Default::default();
-                dict.insert(2, 1_u32);
-                dict.get(2)
-            }
-        );
-
-        run_program_assert_output(&program, "run_test", &[], 1u32.into());
-    }
-
```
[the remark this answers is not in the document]
FrancoGiachetta: My bad, I'll revert that

src/libfuncs/felt252.rs
FrancoGiachetta: Because it's that libfunc that provokes the panic, it returns a `Result`. The sierra generator is meant to run only one libfunc, so, in its happy path, it will nerver panic.

src/values.rs
```diff
@@ -175,6 +175,16 @@ impl Value {
...
 
+        match ty {
+            CoreTypeConcrete::Box(info)
+            | CoreTypeConcrete::NonZero(info)
+            | CoreTypeConcrete::Nullable(info)
+            | CoreTypeConcrete::Snapshot(info) => {
+                return self.to_ptr(arena, registry, &info.ty, find_dict_overrides);
+            }
+            _ => {}
+        }
```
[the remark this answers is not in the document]
FrancoGiachetta: that match below is different, it is a match of `self` which is a `Value`. `ty` is a `CoreTypeConcrete`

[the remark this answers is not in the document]
FrancoGiachetta: That match below is different, it is a match of self which is a Value. ty is a CoreTypeConcrete

src/libfuncs/felt252_dict.rs
FrancoGiachetta: Should be the same as dict, but it seems I forgot to add the assert

src/libfuncs/cast.rs
FrancoGiachetta: Done

src/libfuncs/bytes31.rs
FrancoGiachetta: I think It wouldn't be possible. It would be passing an empty struct which is not expected, it wouldn't even compile.

src/libfuncs/enum.rs
FrancoGiachetta: Mmm, is true that is quite complex. I agree that leaving those in sierra could be better. What are you thoughts on this @gabrielbosio?

src/libfuncs/felt252.rs
azteca1998: The libfunc accepts a NonZero as a denominator, thus passing a zero is not defined since it's not possible. I think we should not test those combinations.

src/utils.rs
```diff
@@ -608,6 +608,30 @@ pub mod test {
             .unwrap()
     }
 
+    pub fn run_sierra_program(program: &Program, args: &[Value]) -> ExecutionResult {
```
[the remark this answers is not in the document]
JulianGCalderon: Could you add a comment explaining the usage of the function? You could mention that it is primarily used with the sierra generator.

src/libfuncs/bytes31.rs
JulianGCalderon: Aah I see. It's because the libfunc fails with Felt::MAX, so it doesn't make sense to add the `to_felt` case, right?

src/libfuncs/felt252.rs
FrancoGiachetta: I can bring back the cairo code for those special cases in a separate test.

src/utils.rs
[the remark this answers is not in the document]
FrancoGiachetta: Done

src/libfuncs/felt252_dict_entry.rs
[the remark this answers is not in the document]
FrancoGiachetta: Done

src/libfuncs/bytes31.rs
FrancoGiachetta: Not only because of that but also because it does not compile, since it does not expect an empty struct as un argument.

src/libfuncs/felt252_dict.rs
FrancoGiachetta: Done

src/values.rs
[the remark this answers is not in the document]
FrancoGiachetta: I'll change that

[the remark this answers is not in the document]
FrancoGiachetta: Done

src/values.rs
```diff
@@ -750,7 +760,11 @@ impl Value {
                 CoreTypeConcrete::Uint64(_) => Self::Uint64(*ptr.cast::<u64>().as_ref()),
                 CoreTypeConcrete::Uint128(_) => Self::Uint128(*ptr.cast::<u128>().as_ref()),
                 CoreTypeConcrete::Uint128MulGuarantee(_) => {
-                    native_panic!("todo: implement uint128mulguarantee from_ptr")
+                    if cfg!(test) {
+                        Self::Null
+                    } else {
+                        native_panic!("todo: implement uint128mulguarantee from_ptr")
+                    }
```
JulianGCalderon, line 767: Why is it only implemented in tests?

src/libfuncs/uint512.rs
```diff
@@ -205,12 +207,18 @@ mod test {
...
+                fields,
+                debug_name: _,
+            } = run_sierra_program(&UINT512_DIVMOD_U256, &[lhs, rhs]).return_value
+            else {
+                panic!("should be a struct");
+            };
+
+            let result = jit_struct!(fields[0].clone(), fields[1].clone());
+
+            assert_eq!(result, jit_struct!(output_u512, output_u256));
```
JulianGCalderon, line 221: Could you also assert the output of all the fields, instead of only the first two?

PR #1075, opened 2025-01-30
Add lockfile support for the `AotContractExecutor`.

azteca1998:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/executor/contract.rs
```diff
@@ -124,162 +125,162 @@ impl BuiltinType {
...
+    }
 
-        let module = native_context.compile(
-            sierra_program,
+    /// Compile and load a program into a shared library.
+    ///
+    /// This function uses a lockfile to support cache sharing between multiple processes. An
+    /// attempt to compile a program while the `output_path` is already locked will result in
+    /// `Ok(None)` being returned. When this happens, the user should wait until the lock is
+    /// released, at which point they can use `AotContractExecutor::from_path` to load it.
```
JulianGCalderon, line 153: Is there a way for the calling process to block until the lock is released?

azteca1998: Yes, either with a busy loop checking if the lockfile still exists or polling the OS for filesystem events, but the later option requires extra dependencies like for example `inotify` and if we did that within native we may end up blocking the user when we shouldn't have, that's why I did it like that instead of blocking. The user can decide whether and how to block, if necessary.

edg-l (approved):

JulianGCalderon (approved):

igaray (dismissed):

igaray (approved):

edg-l (approved):

PR #1078, opened 2025-01-31
Specify sierra version in Contract Executor

### LINKED ISSUE #1072: Provide a way to specifiy the sierra version at AotContractExecutor::new()
Alternatively we can add an argument to specify whether to use the linear gas solver or not

JulianGCalderon:
Adds a new argument to `AotContractExecutor::new` that allows for specifying the sierra version. It is used to correctly build the computation metadata.

Closes #1072

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/executor/contract.rs
+++ b/src/executor/contract.rs
@@ -57,8 +57,10 @@ use cairo_lang_sierra::{
     ids::FunctionId,
     program::Program,
 };
-use cairo_lang_starknet_classes::casm_contract_class::ENTRY_POINT_COST;
 use cairo_lang_starknet_classes::contract_class::ContractEntryPoints;
+use cairo_lang_starknet_classes::{
+    casm_contract_class::ENTRY_POINT_COST, compiler_version::VersionId,
+};
 use educe::Educe;
 use libloading::Library;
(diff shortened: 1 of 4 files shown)


src/executor/contract.rs
```diff
@@ -150,6 +153,7 @@ impl AotContractExecutor {
             used_function_ids.insert(entry.function_idx as u64);
         }
 
+        let no_eq_solver = sierra_version.minor >= 4;
```
edg-l, line 156: i think we should check the major too

edg-l (approved):

src/executor/contract.rs
azteca1998: +1, probably with something like this
```suggestion
        let no_eq_solver = match sierra_version.major.cmp(&1) {
            Ordering::Less => false,
            Ordering::Eq => sierra_version.minor >= 4,
            Ordering::Greater => true,
        };
```

JulianGCalderon: Done

azteca1998 (approved):

edg-l (approved):

PR #1100, opened 2025-02-14
Refactor arrays to avoid clones according to the cost model.

azteca1998:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


edg-l (approved):

src/libfuncs/array.rs
```diff
@@ -749,159 +619,110 @@ pub fn build_append<'ctx, 'this>(
...
+                        realloc_len,
+                        location,
+                    )?)?;
+                    let array_ptr = block.gep(
+                        context,
+                        location,
+                        array_ptr,
+                        &[GepIndex::Const(data_prefix_size as i32)],
+                        IntegerType::new(context, 8).into(),
+                    )?;
```
FrancoGiachetta, line 662: I'm a little confused here. What's the difference between this `array_ptr` and the one above (line 642) whose `GepIndex` operation has a negative value of `data_prefix_size`.

src/libfuncs/array.rs
```diff
@@ -1581,41 +1091,24 @@ pub fn build_get<'ctx, 'this>(
     Ok(())
 }
 
-/// Generate MLIR operations for the `array_slice` libfunc.
+/// Generate MLIR operations for the `array_len` libfunc.
 pub fn build_slice<'ctx, 'this>(
```
FrancoGiachetta, line 1095: I think there was a mistake here and it should array_slice, right?

src/libfuncs/array.rs
azteca1998: The input array_ptr (obtained from the realloc) is the allocation pointer. The offset pointer (result of the gep) is the pointer that will be stored in the array, which points to the first element (after the reference counter and max_len prefix).

The same process applies, but in reverse at line 642. Realloc needs the pointer returned by the original allocation, so we need to undo the offset (thus the negative) to obtain the original pointer and realloc.

src/libfuncs/array.rs
azteca1998: Yes, my mistake.

src/types/array.rs
```diff
@@ -405,6 +403,17 @@ fn build_drop<'ctx>(
...
+    get_integer_layout(32)
+        .extend(get_integer_layout(32))
+        .unwrap()
+        .0
+        .align_to(layout.align())
+        .unwrap()
+        .pad_to_align()
+        .size()
+}
+
```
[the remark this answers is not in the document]
azteca1998: Should never fail, so I don't know if using expect makes sense. Would a comment explaining that it's impossible to trigger them work?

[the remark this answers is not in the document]
JulianGCalderon: > Would a comment explaining that it's impossible to trigger them work?

That should be fine, but isn't an unwrap + comment the same as an expect?

src/executor/contract.rs
```diff
@@ -520,62 +527,48 @@ impl AotContractExecutor {
         let tag = *unsafe { enum_ptr.cast::<u8>().as_ref() } as usize;
         let tag = tag & 0x01; // Filter out bits that are not part of the enum's tag.
 
-        // layout of both enum variants, both are a array of felts
```
JulianGCalderon, line 523: Is this comment no longer valid? If it is, could you add it back?

src/types/array.rs
JulianGCalderon, line 1: Could you update this file's top-level documentation with the new layout?

src/types/array.rs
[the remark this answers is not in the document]
JulianGCalderon: > The integers are:
> - Reference counter: literally the number of references to the allocation.
> - Max length: The number of elements present in the allocation (not necessarily the length array/span being accessed, but the whole allocation). It is used to know how many elements to drop when freeing the allocation.

Could you add that text to the code? Maybe to lines 17-20 in this file.

JulianGCalderon (commented):
The PR looks good to me, although I think the documentation could be improved. I left some comments with ideas on how to improve it.

Also, the `build_pop` function (and overall the whole file) is much easier to understand now, nice work.

JulianGCalderon (approved):

gabrielbosio (approved):

PR #1103, opened 2025-02-18
Remove panics

JulianGCalderon:
Removes some panics in Cairo Native

### DIFF
--- a/src/libfuncs/array.rs
+++ b/src/libfuncs/array.rs
@@ -712,7 +712,7 @@ pub fn build_append<'ctx, 'this>(
 
             metadata
                 .get::<DropOverridesMeta>()
-                .unwrap()
+                .to_native_assert_error("drop overrides meta should exist")?
                 .invoke_override(
                     context,
                     &block,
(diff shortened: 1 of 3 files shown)


src/libfuncs/array.rs
```diff
@@ -1341,7 +1341,7 @@ fn build_pop<'ctx, 'this, const CONSUME: bool, const REVERSE: bool>(
 
             metadata
                 .get::<DropOverridesMeta>()
-                .unwrap()
+                .to_native_assert_error("drop overrides meta should exist")?
```
FrancoGiachetta, line 1344: Why not using `?` operator in cases like this one? Since most of the functions return a `Result`

edg-l (approved):

src/libfuncs/array.rs
JulianGCalderon: `get` does not return a Result, so we can't use the `?` operator.

FrancoGiachetta (approved):

azteca1998 (approved):

PR #1115, opened 2025-02-25
Improves documentation on unreachable panics

JulianGCalderon:
Change unwrap to expect in some unreachable panics, explaining why they will never fail.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/types/array.rs
+++ b/src/types/array.rs
@@ -406,10 +406,10 @@ fn build_drop<'ctx>(
 pub fn calc_data_prefix_offset(layout: Layout) -> usize {
     get_integer_layout(32)
         .extend(get_integer_layout(32))
-        .expect("creating a layout of two i32 will never fail")
+        .expect("creating a layout of two i32 should never fail")
         .0
         .align_to(layout.align())
-        .expect("only fails if size is greater than approximately ISIZE::MAX")
+        .expect("layout size rounded up to the next multiple of layout alignment should never be greater than ISIZE::MAX")
         .pad_to_align()
         .size()
 }
--- a/src/types/array.rs
+++ b/src/types/array.rs
@@ -406,10 +406,10 @@ fn build_drop<'ctx>(
 pub fn calc_data_prefix_offset(layout: Layout) -> usize {
     get_integer_layout(32)
         .extend(get_integer_layout(32))
-        .unwrap()
+        .expect("creating a layout of two i32 will never fail")
         .0
         .align_to(layout.align())
-        .unwrap()
+        .expect("only fails if size is greater than approximately ISIZE::MAX")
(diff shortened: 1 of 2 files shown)


src/types/array.rs
```diff
@@ -406,10 +406,10 @@ fn build_drop<'ctx>(
 pub fn calc_data_prefix_offset(layout: Layout) -> usize {
     get_integer_layout(32)
         .extend(get_integer_layout(32))
-        .unwrap()
+        .expect("creating a layout of two i32 will never fail")
```
gabrielbosio, line 409: Following the Recommended Messaging Style from [Rust docs](https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style):

```suggestion
        .expect("creating a layout of two i32 should never fail")
```

src/types/array.rs
```diff
@@ -406,10 +406,10 @@ fn build_drop<'ctx>(
...
         .0
         .align_to(layout.align())
-        .unwrap()
+        .expect("only fails if size is greater than approximately ISIZE::MAX")
```
gabrielbosio, line 412: I suggest to replace `approximately` with something a bit more specific like:

```suggestion
        .expect("only fails if layout size rounded up to the next multiple of layout alignment is greater than ISIZE::MAX")
```

You can even rename it to:

```suggestion
        .expect("layout size rounded up to the next multiple of layout alignment should never be greater than ISIZE::MAX")
```

src/utils.rs
```diff
@@ -176,7 +176,7 @@ pub fn get_integer_layout(width: u32) -> Layout {
     } else {
         // According to the docs this should never return an error.
         Layout::from_size_align((width as usize).next_multiple_of(8) >> 3, 16)
-            .expect("only fails if size is greater than ISIZE::MAX")
+            .expect("only fails if size is greater than approximately ISIZE::MAX")
```
gabrielbosio, line 179: I suggest to replace `approximately` with something a bit more specific like:

```suggestion
        .expect("only fails if size rounded up to the next multiple of 16 is greater than ISIZE::MAX")
```

You can even rename it to:

```suggestion
        .expect("layout size rounded up to the next multiple of 16 should never be greater than ISIZE::MAX")
```

gabrielbosio (changes requested):

src/types/array.rs
JulianGCalderon: Done

src/utils.rs
JulianGCalderon: Done

gabrielbosio (approved):

FrancoGiachetta (approved):

PR #1116, opened 2025-02-25
Remove panics from libfuncs directory

JulianGCalderon:
This PR removes all panics in the libfuncs directory, by replacing it with native asserts.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/starknet.rs
+++ b/src/libfuncs/starknet.rs
@@ -145,7 +145,9 @@ pub fn build<'ctx, 'this>(
         }
         #[cfg(not(feature = "with-cheatcode"))]
         StarkNetConcreteLibfunc::Testing(TestingConcreteLibfunc::Cheatcode(_)) => {
-            unimplemented!("feature 'with-cheatcode' is required to compile with cheatcode syscall")
+            crate::native_panic!(
+                "feature 'with-cheatcode' is required to compile with cheatcode syscall"
+            )
         }
     }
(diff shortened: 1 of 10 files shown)


FrancoGiachetta (dismissed):

src/libfuncs/starknet.rs
```diff
@@ -145,7 +145,9 @@ pub fn build<'ctx, 'this>(
         }
         #[cfg(not(feature = "with-cheatcode"))]
         StarkNetConcreteLibfunc::Testing(TestingConcreteLibfunc::Cheatcode(_)) => {
-            unimplemented!("feature 'with-cheatcode' is required to compile with cheatcode syscall")
+            crate::native_panic!(
```
gabrielbosio, line 148: Why `crate::native_panic!` instead of `native_panic!`?

JulianGCalderon: As its only required when NOT using `with-cheatcode` feature. Adding an import would give unused warnings when executing with all features. I could gate the import under the flag, but I though it was simpler this way.

gabrielbosio: Ok, makes sense

gabrielbosio (approved):

FrancoGiachetta (approved):

gabrielbosio (approved):

PR #1125, opened 2025-02-28
Use SierraGenerator for testing (part 1)

FrancoGiachetta:
This PR changes some tests to use Sierra code, rather than Cairo code.

This has the following advantages:
- Reduces test execution time by omitting compilation.
- We ensure that the test is not modified by changes in the compiler

Related to #1152

## Checklist
- [x] Linked to Github Issue


edg-l (approved):

azteca1998 (dismissed):

src/values.rs
```diff
@@ -169,6 +169,16 @@ impl Value {
...
+        match ty {
+            CoreTypeConcrete::Box(info)
+            | CoreTypeConcrete::NonZero(info)
+            | CoreTypeConcrete::Nullable(info)
+            | CoreTypeConcrete::Snapshot(info) => {
+                return self.to_ptr(arena, registry, &info.ty, find_dict_drop_override);
+            }
+            _ => {}
+        }
+
```
JulianGCalderon, line 181: I don't fully understand this: if we try to call `to_ptr` with a Box<uint32>, then this function would return a pointer to an uint32, and not a double pointer to the uint32 (which is the correct type), right?

src/libfuncs/bytes31.rs
```diff
@@ -162,39 +162,53 @@ pub fn build_from_felt252<'ctx, 'this>(
...
-            &BYTES31_ROUNDTRIP,
-            "run_test",
-            &[Felt::MAX.into()],
-            jit_panic!(Felt::from_bytes_be_slice(b"Option::unwrap failed.")),
-        );
+        let Value::Enum {
+            tag: _,
+            value,
+            debug_name: _,
+        } = run_sierra_program(&BYTES31_TRY_FROM_FELT, &[Felt::from(2).into()]).return_value
```
JulianGCalderon, line 202: Here you could use the shorter syntax: `Value::Enum { value.. }`

src/libfuncs/bounded_int.rs
```diff
@@ -817,134 +817,117 @@ fn build_wrap_non_zero<'ctx, 'this>(
...
+        let Value::Enum {
+            tag: 1,
+            value,
+            debug_name: _,
+        } = run_sierra_program(&program_trim, &[Value::Sint8(1)]).return_value
+        else {
+            panic!("should be OptionRev::Some");
+        };
+
+        let Value::BoundedInt { value, range: _ } = *value else {
```
JulianGCalderon, line 857: Could we assert that the range now is shorter?

src/values.rs
```diff
@@ -746,9 +756,7 @@ impl Value {
                 CoreTypeConcrete::Uint32(_) => Self::Uint32(*ptr.cast::<u32>().as_ref()),
                 CoreTypeConcrete::Uint64(_) => Self::Uint64(*ptr.cast::<u64>().as_ref()),
                 CoreTypeConcrete::Uint128(_) => Self::Uint128(*ptr.cast::<u128>().as_ref()),
-                CoreTypeConcrete::Uint128MulGuarantee(_) => {
-                    native_panic!("todo: implement uint128mulguarantee from_ptr")
-                }
+                CoreTypeConcrete::Uint128MulGuarantee(_) => Self::Null,
```
JulianGCalderon, line 759: As `Null` is only for nullables, i'd add a new variant `Unit` that could be used in these situations.

src/libfuncs/cast.rs
```diff
@@ -391,299 +391,176 @@ pub fn build_upcast<'ctx, 'this>(
...
+        program::GenericArg,
+    };
+
+    macro_rules! cast {
+        ($from:ty, $to:ty, $is_up_cast:expr) => {
+            if $is_up_cast {
+                let mut generator = SierraGenerator::<UpcastLibfunc>::default();
+
+                let from_ty = generator.push_type_declaration::<$from>(&[]).clone();
+                let to_ty = generator.push_type_declaration::<$to>(&[]).clone();
```
JulianGCalderon, line 418: Could this macro be changed to a function that receives the input types as generic parameters?

FrancoGiachetta: I would do that, the problem here is that rust macros do not allow `let` expressions unless there are inside an `if` or `while` statement. So I would still need a boolean to differentiate.

JulianGCalderon: I meant that instead of a macro, we could have a generic function (I think it should be pretty similar and easier to understand). The function could also have a generic boolean argument for the `is_upcast`.

src/libfuncs/bytes31.rs
FrancoGiachetta: Done

src/libfuncs/bounded_int.rs
FrancoGiachetta: Done

src/libfuncs/cast.rs
FrancoGiachetta: Oh, I see. I'll try that then

FrancoGiachetta: For some reason using generics does not seem to be a solution. Doing it this way makes SierraGenerator is unable to generate the program, whereas with a macro works without issues.

This is the implementation I'm using:
```
fn cast<TO: NamedType, FROM: NamedType, const IS_UPCAST: bool>() -> Program {
        if IS_UPCAST {
            let mut generator = SierraGenerator::<UpcastLibfunc>::default();
            
            let from_ty = generator.push_type_declaration::<FROM>(&[]).clone();
            let to_ty = generator.push_type_declaration::<TO>(&[]).clone();
            
            ...
        } else {
            let mut generator = SierraGenerator::<DowncastLibfunc>::default();
            
            let from_ty = generator.push_type_declaration::<FROM>(&[]).clone();
            let to_ty = generator.push_type_declaration::<TO>(&[]).clone();
            
            ...
        }
    }
```
Do you spot any errors? If not I don't think this could be a solution.

FrancoGiachetta: For downcast
```
thread 'libfuncs::cast::test::downcast' panicked at src/utils.rs:633:14:
called `Result::unwrap()` on an `Err` value: NativeAssert(NativeAssertError { msg: "todo: abi argument unimplemented for (Uint128(340282366920938463463374607431768211455), ConcreteTypeId { id: 0, debug_name: None })", info: Location(Location { file: "src/error.rs", line: 163, col: 47 }) })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
For upcast
```
thread 'libfuncs::cast::test::upcast' panicked at src/utils/sierra_gen.rs:80:14:
called `Result::unwrap()` on an `Err` value: UnsupportedGenericArg
```

Looks like it is receiving the wrong arguments, but they are the same as with the macro.

JulianGCalderon: Mmh, sounds odd. If you want I can take a closer look, could you push the changes somewhere (maybe to another branch)?

src/values.rs
azteca1998: At a glance, the match should work for `NonZero` and `Snapshot`, but should fail for `Box` and `Nullable`.

Both `NonZero` and `Snapshot` have the same internal representation as the original type, but `Box` and `Nullable` are heap pointers, so we need to allocate and initialize them.
Additionally, `Nullable` should handle the `Value::Null` variant.

FrancoGiachetta: mmm, both box and nullable tests are passing without any failures. Perhaps this is because the values are not really pointers since they are passed as an arguments through the test. Does this have an impact outside of tests? If not maybe I should leave it only for tests.

JulianGCalderon: Right now It will only be used for tests, but its better to implement it properly in case we need it later, right? If not we could be leaving bugs that show up sooner or later.

FrancoGiachetta:
This PR makes testing libfuncs quite complicated and its objective can be achieved by other, simpler, means. So, it has no meaning anymore.

PR #1129, opened 2025-02-28
Use SierraGenerator for testing (part 5) 

FrancoGiachetta:
This PR changes some tests to use Sierra code, rather than Cairo code.

This has the following advantages:
- Reduces test execution time by omitting compilation.
- We ensure that the test is not modified by changes in the compiler

Related to #1152

## Checklist
- [x] Linked to Github Issue


FrancoGiachetta:
This PR makes testing libfuncs quite complicated and its objective can be achieved by other, simpler, means. So, it has no meaning anymore.

PR #1141, opened 2025-03-20
Add `meta_tx_v0` syscall.

azteca1998:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


tests/tests/starknet/syscalls.rs
```diff
@@ -398,6 +398,47 @@ impl StarknetSyscallHandler for SyscallHandler {
...
+                "3358892263739032253767642605669710712087178958719188919195252597609334880396",
+            )
+            .unwrap(),
+            Felt::from_dec_str(
+                "1104291043781086177955655234103730593173963850634630109574183288837411031513",
+            )
+            .unwrap(),
+            Felt::from_dec_str(
+                "3346377229881115874907650557159666001431249650068516742483979624047277128413",
+            )
```
gabrielbosio, line 437: Why this function returns these felts?

src/libfuncs/starknet.rs
```diff
@@ -2124,6 +2123,221 @@ pub fn build_library_call<'ctx, 'this>(
...
+    //     entry_point_selector: felt252,
+    //     calldata: Span<felt252>,
+    // ) -> SyscallResult<Span<felt252>> implicits(GasBuiltin, System) nopanic;
+    //
+    // extern fn meta_tx_v0_syscall(
+    //     address: ContractAddress,
+    //     entry_point_selector: felt252,
+    //     calldata: Span<felt252>,
+    //     signature: Span<felt252>,
+    // ) -> SyscallResult<Span<felt252>> implicits(GasBuiltin, System) nopanic;
```
JulianGCalderon, line 2146: Could you move this the cairo signature of the syscall to the function doc comment, and a description of what the syscall does?

Maybe we could reuse the corelib documentation?
```cairo
/// Invokes the given entry point as a v0 meta transaction.
///
/// * The signature is replaced with the given signature.
/// * The caller is the OS (address 0).
/// * The transaction version is replaced by 0.
/// * The transaction hash is replaced by the corresponding version-0 transaction hash.
///
/// The changes apply to the called contract and the inner contracts it calls.
///
/// NOTE: This syscall should only be used to allow support for old version-0 bound accounts,
/// and should not be used for other purposes.
extern fn meta_tx_v0_syscall(
    address: ContractAddress,
    entry_point_selector: felt252,
    calldata: Span<felt252>,
    signature: Span<felt252>,
) -> starknet::SyscallResult<Span<felt252>> implicits(GasBuiltin, System) nopanic;
```

gabrielbosio (dismissed):

src/libfuncs/starknet.rs
JulianGCalderon: Moves this to a new issue: https://github.com/lambdaclass/cairo_native/issues/1161

JulianGCalderon (approved):
LGTM

PR #1171, opened 2025-04-08
Fix into_u96_guarantee not handling other types

JulianGCalderon:
This PR fixes `into_u96_guarantee` libfunc so that it handles:

- BoundedInts with non zero lower range
- Smaller data types (like u8)

It used to be a noop because we though it always received an u96.

This PR fixes a bug in some contracts, caused by using an `u8` as input to this libfunc:
- `0x1aa0ae9d42edb490e438b161504f01b885e487f73d146a4d10baf78bb3f155`
- `0xc9e956f41095d8b61d7254e6955521121527ffbaad87c6475b71247aa009d4`

Note: Independly from the bug fix, i added a `#[track_caller]` attribute that was missing. It does not change the fix.

### DIFF
--- a/src/libfuncs/circuit.rs
+++ b/src/libfuncs/circuit.rs
@@ -1130,8 +1130,8 @@ mod test {
         values::Value,
     };
     use cairo_lang_sierra::extensions::utils::Range;
-    use num_bigint::BigUint;
-    use num_traits::Num;
+    use num_bigint::{BigInt, BigUint};
+    use num_traits::{Num, One};
     use starknet_types_core::felt::Felt;
 
     fn u384(limbs: [&str; 4]) -> Value {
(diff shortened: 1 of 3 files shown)


FrancoGiachetta (approved):

src/libfuncs/circuit.rs
```diff
@@ -1083,6 +1084,52 @@ fn build_array_slice<'ctx>(
...
+
+    let src_range = src_ty.integer_range(registry)?;
+
+    // We expect the input value to be unsigned, but we check it just in case.
+    if src_range.lower.is_negative() {
+        native_panic!("into_u96_guarantee expects an unsigned integer")
+    }
+
+    // Extend the input value to an u96
+    let mut dst = entry.extui(src, IntegerType::new(context, 96).into(), location)?;
```
azteca1998, line 1115: What if the range is larger than an `u96`? The generated MLIR would still be invalid.

JulianGCalderon: Compiler doesn't allow to specify a size greater than that value.

And if that ever happens, MLIR will catch the error instead of allowing it to run, right?

azteca1998 (approved):

PR #1174, opened 2025-04-10
Refactor DebugUtils with support for both AOT and JIT executors

JulianGCalderon:
Debug utils (prints, breakpoints) used to work only for JIT executors. This is because the implementation of this functions was actually in Rust, and we loaded the symbols directly in the JIT engine.

This PR refactors the debug utils to work both for AOT and JIT executors. The approach taken is similar (identical) to how we manage the Runtime. See: https://github.com/lambdaclass/cairo_native/pull/1051

TLDR: We declare globals for each rust function we intend to call, and when loading a contract we set the global values to point to the address of the rust function.


FrancoGiachetta (approved):

FrancoGiachetta (approved):

src/metadata/debug_utils.rs
```diff
@@ -118,44 +118,117 @@ enum DebugBinding {
     DumpMemRegion,
 }
 
+impl DebugBinding {
+    const fn symbol(self) -> &'static str {
+        match self {
+            DebugBinding::BreakpointMarker => "cairo_native__debug__breakpoint_marker_impl",
```
gabrielbosio, line 124: Does using `cairo_native__debug__breakpoint_marker_impl` instead of `__debug__breakpoint_marker` improve legibility?
The same for the rest of the enum variants

JulianGCalderon: The reason I renamed it is to keep consistency with the other runtime symbols we define:

https://github.com/lambdaclass/cairo_native/blob/9815958c6eb70dbaaabffd4df1de6920c8dc0b39/src/metadata/runtime_bindings.rs#L62-L66

The name is purely internal, and not referenced in any other place. As long as the chosen name doesn't clash with any other symbol, then there should be no problem.

I think that by prefixing the symbol name with `cairo_native`, we ensure that the name will not clash with any other linked library.

Issue #1186, opened 2025-04-21
[Sierra-Emu] Coverage tracking issue


FrancoGiachetta:
We should increase the coverage of `sierra-emu`. Before doing so, we need to implement the `todo`s  left around the project:

- [x] src/debug.rs 
- [x] src/main.rs
- [ ] src/value.rs
- [ ] src/vm/array.rs
- [ ] src/vm/bounded_int.rs
- [x] src/vm/cast.rs
- [ ] src/vm/circuit.rs
- [x] src/vm/const.rs
- [ ] src/vm/ec.rs
- [ ] src/vm/felt252.rs
- [x] src/vm/gas.rs
- [ ] src/vm/int128.rs
- [x] src/vm/starknet.rs
- [x]  src/vm/nullable.rs
- [ ] src/vm/uint8.rs
- [ ] src/vm/uint16.rs
- [ ] src/vm/uint32.rs
- [ ] src/vm/uint64.rs
- [ ] src/vm.rs

FrancoGiachetta:
Finished with https://github.com/lambdaclass/cairo_native/pull/1188

PR #1188, opened 2025-04-22
[Sierra-Emu] Run corelib's tests

FrancoGiachetta:
This PR adds a test to run every corelib's tests. This is to increase the coverage of sierra-emu which is currently quite low. Additionally, I've made a little refactor inside main, moving the program execution logic to method inside lib.rs. This is mainly for a more comfortable usage.
This test is meant to be used in a ci workflow eventually.

Refers to #1186 

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
- [X] Integration tests added.


debug_utils/sierra-emu/src/args.rs
```diff
@@ -13,20 +14,3 @@ pub struct CmdArgs {
...
-impl FromStr for EntryPoint {
-    type Err = ParseIntError;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        Ok(match s.chars().next() {
-            Some(x) if x.is_numeric() => Self::Number(s.parse()?),
-            _ => Self::String(s.to_string()),
-        })
-    }
-}
```
JulianGCalderon, line 32: Why was this moved to another file?

debug_utils/sierra-emu/src/value.rs
```diff
@@ -32,6 +34,8 @@ pub enum Value {
         self_ty: ConcreteTypeId,
         index: usize,
         payload: Box<Self>,
+        #[educe(PartialEq(ignore))]
+        debug_name: Option<String>,
```
JulianGCalderon, line 38: Why is it necessary to add this field? Isn't it contained in `self_ty`?

debug_utils/sierra-emu/tests/corelib.rs
```diff
@@ -0,0 +1,241 @@
...
+use sierra_emu::{run_program, EntryPoint, ProgramTrace, Value};
+use tracing::info;
+use tracing_test::traced_test;
+
+mod common;
+
+#[traced_test]
+#[test]
+#[ignore = "sierra-emu is not fully implemented yet"]
+fn test_corelib() {
```
JulianGCalderon, line 31: I think we should not ignore the tests.

We can ignore the error in the CI instead, that way we can:
1. Assert that this test works as intented.
2. Track progress on the sierra emu implementation.

debug_utils/sierra-emu/tests/common/mod.rs
```diff
@@ -0,0 +1,89 @@
+use sierra_emu::Value;
+use starknet_crypto::Felt;
+
+/// Convert a Value to a felt.
+pub fn jitvalue_to_felt(value: &Value) -> Vec<Felt> {
```
JulianGCalderon, line 5: Its not a `jitvalue`, but a sierra emu value:
```suggestion
/// Convert a Value to a felt.
pub fn value_to_felt(value: &Value) -> Vec<Felt> {
```

debug_utils/sierra-emu/src/args.rs
FrancoGiachetta: It was moved to lib.rs. So I could use it inside `run_program`

debug_utils/sierra-emu/src/value.rs
FrancoGiachetta: Yes, it is used in `tests::corelib::trace_to_run_result` to check if an `Enum` refers to a panic.

debug_utils/sierra-emu/tests/corelib.rs
FrancoGiachetta: I would great, but I don't know how to ignore it from the ci

debug_utils/sierra-emu/tests/common/mod.rs
FrancoGiachetta: My bad

FrancoGiachetta: Done

debug_utils/sierra-emu/src/value.rs
JulianGCalderon: If its already contained in `self_ty`, can't you access it from there instead of adding a new field?

debug_utils/sierra-emu/tests/corelib.rs
FrancoGiachetta: I could ignore it and run it separately. That way we can still run make test during the build check workflow and then ignore the result of this test

debug_utils/sierra-emu/src/value.rs
FrancoGiachetta: mmm, I guess I could. I'll remove it then

FrancoGiachetta: Done

debug_utils/sierra-emu/tests/corelib.rs
FrancoGiachetta: I did that for now

.github/workflows/ci.yml
```diff
@@ -359,3 +359,7 @@ jobs:
     - name: Run tests
       working-directory: ./debug_utils/sierra-emu
       run: make test
+    - name: Run corelib's tests
```
gabrielbosio, line 362: I would add here something to make more visible that this job is in progress. Maybe:

```suggestion
    - name: Run corelib's tests (WIP)
```

.github/workflows/ci.yml
```diff
@@ -341,7 +341,7 @@ jobs:
           df -h
       - name: build image
         run: docker build .
-  
+
```
JulianGCalderon, line 344: Could you revert this change?

debug_utils/sierra-emu/src/dump.rs
```diff
@@ -80,6 +93,7 @@ impl ContractExecutionResult {
                     self_ty: _,
                     index,
                     payload,
+                    ..
```
JulianGCalderon, line 96: I think this line is no longer required. Idem for all the other `..`

debug_utils/sierra-emu/src/vm/bool.rs
```diff
@@ -48,9 +48,9 @@ pub fn eval_and(
     EvalAction::NormalBranch(
         0,
         smallvec![Value::Enum {
-            self_ty,
+            self_ty: self_ty.clone(),
```
JulianGCalderon, line 51: I think the clone is not necessary. Idem for all the other `self_ty.clone()`s

debug_utils/sierra-emu/tests/corelib.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    let compiled = compile_tests(
+        &db,
+        test_config,
+        test_crate_ids,
+        diag_reporter,
+        Some(&filtered_tests),
+    );
+
+    run_tests(compiled);
```
JulianGCalderon, line 75: Is there a way to print a report like Cairo Native does? That way we can clearly see which tests are failing and which are not.

.github/workflows/ci.yml
FrancoGiachetta: Yes I agree, done.

.github/workflows/ci.yml
FrancoGiachetta: I don't understand that diff, if you check it in the file, there's no change at all compared to main.

debug_utils/sierra-emu/src/dump.rs
gabrielbosio, line 96: Why is this needed?

gabrielbosio (changes requested):

debug_utils/sierra-emu/tests/corelib.rs
```diff
@@ -0,0 +1,226 @@
...
+                EntryPoint::String(name.clone()),
+                vec![],
+                u64::MAX,
+            );
+            let run_result = trace_to_run_result(trace);
+
+            *success &= assert_test_expectation(name, test.expectation, run_result);
+        });
+
+    assert!(success, "Some tests from the corelib have failed");
```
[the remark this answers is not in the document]
FrancoGiachetta: You are right, i'll change it

debug_utils/sierra-emu/src/vm/bool.rs
FrancoGiachetta: Removed

debug_utils/sierra-emu/tests/corelib.rs
```diff
@@ -0,0 +1,226 @@
...
+            .for_each(|(test, case)| {
+                if should_skip_test(test) {
+                    case.ignored = true
+                }
+            });
+
+        compiled
+    } else {
+        compiled
+    }
```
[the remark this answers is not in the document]
FrancoGiachetta: Done

debug_utils/sierra-emu/tests/corelib.rs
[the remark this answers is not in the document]
FrancoGiachetta: Done

debug_utils/sierra-emu/src/vm/bool.rs
```diff
@@ -103,9 +103,9 @@ pub fn eval_xor(
...
```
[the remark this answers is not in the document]
FrancoGiachetta: Removed

debug_utils/sierra-emu/src/dump.rs
FrancoGiachetta: Removed

debug_utils/sierra-emu/src/vm/bool.rs
```diff
@@ -134,9 +134,9 @@ pub fn eval_or(
...
```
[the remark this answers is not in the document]
FrancoGiachetta: Removed

.github/workflows/ci.yml
JulianGCalderon: There is a difference, but now that I think of it it doesn't really matter. Main has some trailing whitespace con line 344, while this branch has not. I think you could leave it as is.

debug_utils/sierra-emu/src/vm/bool.rs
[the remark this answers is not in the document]
FrancoGiachetta: Removed

debug_utils/sierra-emu/tests/corelib.rs
JulianGCalderon: The cairo native report prints a summary report at the end:
```
test result: ok. 613 passed; 0 failed; 13 ignored; 0 filtered out;
```
We have the same for the sierra emu? I can't find it.

I think another problem is that the tests panic. This implies that the output of some tests is not shown: https://github.com/lambdaclass/cairo_native/actions/runs/14651275989/job/41117426171#step:7:42. Maybe we could catch the panic so that we can clearly see which tests are missing.

debug_utils/sierra-emu/tests/corelib.rs
```diff
@@ -0,0 +1,225 @@
...
+                program.clone(),
+                EntryPoint::String(name.clone()),
+                vec![],
+                u64::MAX,
+            );
+            let run_result = trace_to_run_result(trace);
+
+            assert_test_expectation(name, test.expectation, run_result)
+        })
+        .all(|r| r);
```
JulianGCalderon, line 99: `all()` and `map()` are short-circuiting; in other words, it will stop processing as soon as it finds a false, given that no matter what else happens, the result will also be false.

Could you add a `collect` to avoid this, and use `all` after it?

JulianGCalderon (changes requested):

debug_utils/sierra-emu/tests/corelib.rs
FrancoGiachetta: I see, I'll change that

FrancoGiachetta: Done

debug_utils/sierra-emu/tests/corelib.rs
FrancoGiachetta: Well, I've tried to catch panics. There a function `catch_unwind` which retrieved the error message from the panic. The problem is that many panics are todos with no message and even if the had it does not tell where it was originated. It would be better the make a refactor and return errors instead of panicking. I'll leave the code since it helps getting the test which have panicked, but some of them won't tell the actual error.

.github/workflows/ci.yml
gabrielbosio: It seems [this commit](https://github.com/lambdaclass/cairo_native/commit/48fb545ea0021f6235f669811173037568d7ad55) reverted the suggested change.

JulianGCalderon (approved):
LGTM

.github/workflows/ci.yml
FrancoGiachetta: Oh my bad. Changing it.

debug_utils/sierra-emu/Makefile
```diff
@@ -13,11 +13,12 @@ endif
 
 usage:
 	@echo "Usage:"
-	@echo "    deps:		 Installs the necesarry dependencies."
-	@echo "    build:        Builds the cairo-native library and binaries."
-	@echo "    check:        Checks format and lints."
-	@echo "    test:         Runs all tests."
-	@echo "    clean:        Cleans the built artifacts."
+	@echo "    deps:		    Installs the necesarry dependencies."
```
JulianGCalderon, line 16:
```suggestion
	@echo "    deps:         Installs the necesarry dependencies."
```

JulianGCalderon (changes requested):
The corelib test is being ignored:
> test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.00s

Maybe the name is not right.

FrancoGiachetta:
> The corelib test is being ignored:
> 
> > test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.00s
> 
> Maybe the name is not right.

Everything working now

JulianGCalderon (approved):
LGTM

PR #1199, opened 2025-04-29
Add Trace Dump feature

JulianGCalderon:
This PR adds the trace dump feature to main. I encapsulated everything under the feature flag `with-trace-dump`, so that we can safely merge it.

The main structure is `ProgramTrace`, which contains a list of `StateDump`s, one for each instruction:
```rust
pub struct ProgramTrace {
    pub states: Vec<StateDump>,
}
pub struct StateDump {
    pub statement_idx: StatementIdx,
    pub items: BTreeMap<u64, Value>,
}
```

Right before every statement, in `compiler.rs`, we call `build_state_snapshot` from `utils/trace_dump.rs`.

When calling `build_state_snapshot`, we build code that iterates each known variable and pushes a `state dump` entry to the current `trace dump`. To make implementation easier, we use the metadata `TraceDumpMeta` to call runtime functions in Rust that contain the actual logic of building the state dump.

The trace dump is stored on a global static variable:
```rust
pub static TRACE_DUMP: LazyLock<Mutex<HashMap<u64, TraceDump>>> =
    LazyLock::new(|| Mutex::new(HashMap::new()));
```
The key refers to the trace ID, unique to each program/contract execution. This allows for taking a trace dump of a contract with inner calls.

I added some integration tests to compare the trace dump with the one generated with the sierra emulator. 

Also, I substracted the initial required gas on sierra emu execution, so that the generated trace dump is the same.

## Checklist


src/metadata/trace_dump.rs
```diff
@@ -0,0 +1,783 @@
...
+        pub fn new(registry: ProgramRegistry<CoreType, CoreLibfunc>) -> Self {
+            Self {
+                trace: ProgramTrace::default(),
+                state: OrderedHashMap::default(),
+                registry,
+            }
+        }
+    }
+
+    pub unsafe extern "C" fn add_to_state(
```
FrancoGiachetta, line 253: Could this method be renamed to something like `add_variable_to_state`? The name is a little similar to `push_state` and quite confusing. Or perhaps we could add a comment in both methods explaining their purpose.

JulianGCalderon: Done

src/utils/trace_dump.rs
```diff
@@ -0,0 +1,225 @@
...
+//!    See `TRACE_DUMP`.
+//!
+//! 2. We must store somewhere the ID of the current trace dump, and update it
+//!    acordingly when switching between contract executors
+//!
+//! Both these elements must be properly setup before running the executor. See
+//! `cairo-native-run` for an example on how to do it. You can also check on
+//! this file's integration tests.
+//!
+//! When executing starknet contracts, then the trace id must be set right
```
gabrielbosio, line 63:
```suggestion
//! When executing starknet contracts, the trace id must be set right
```

src/bin/cairo-native-run.rs
```diff
@@ -11,7 +11,7 @@ use cairo_native::{
     starknet_stub::StubSyscallHandler,
 };
 use clap::{Parser, ValueEnum};
-use std::path::PathBuf;
+use std::{fs::File, path::PathBuf};
```
gabrielbosio, line 14:
```suggestion
use std::path::PathBuf;

#[cfg(feature = "with-trace-dump")]
use std::fs::File;
```

src/bin/cairo-native-run.rs
gabrielbosio, line 1: It would be nice if the documentation of this binary is updated, including:

- Usage
- README

src/bin/cairo-native-run.rs
JulianGCalderon: I solved by using the full qualified path instead

src/bin/cairo-native-run.rs
JulianGCalderon: I updated the cli args documentation, and added a section in the `debugging` document, explaining how to invoke it: https://github.com/lambdaclass/cairo_native/pull/1199/commits/2cd752797ad27743c1d6a5ec235101ecc3a4b0f8

gabrielbosio (dismissed):

FrancoGiachetta (approved):

gabrielbosio (approved):

PR #1202, opened 2025-04-30
Minor improvements to Sierra Emu crate

JulianGCalderon:
This PR makes some improvements to the sierra emu crate:

- Move some general dependencies to the workspace, to make updating versiones easier.
- Simplify Makefile, and relying instead on reusing the root Makefile.
- Add more documentation on how to use Sierra Emu as a binary
- Add examples on how to user Sierra Emu as a library

### DIFF
--- a/debug_utils/sierra-emu/Makefile
+++ b/debug_utils/sierra-emu/Makefile
@@ -1,6 +1,6 @@
 usage:
 	@echo "Usage:"
-	@echo "    corelib:	        Install the corelib."
+	@echo "    corelib:         Install the corelib."
 	@echo "    build:           Builds the sierra emu."
 	@echo "    check:           Checks format and lints."
 	@echo "    test:            Runs all tests."


FrancoGiachetta (approved):

debug_utils/sierra-emu/examples/program.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    // Execute the virtual machine.
+    let syscall_handler = &mut StubSyscallHandler::default();
+    let trace = vm.run_with_trace(syscall_handler);
+
+    // Obtain execution result from last frame
+    let result =
+        ContractExecutionResult::from_trace(&trace).ok_or("failed to obtain execution result")?;
+
+    println!("{result:?}");
```
FrancoGiachetta, line 42: The last merge adds a function, exposed through lib.rs, `run_program` which wraps this and makes it easier to run a program. perhaps you can use that one instead. If you do, keep in mind that does not handle contracts but it should be easy to create as well

FrancoGiachetta (dismissed):

debug_utils/sierra-emu/examples/program.rs
JulianGCalderon: I think I'd prefer to do that in a different pull request. This PR does not modify any code, and I think its better to leave it that way.

Using `run_program` would imply refactoring it a bit, so that it can receive `Vec<Value>` instead of `Vec<String>`. Maybe we can divide it into two functions, one that parses CLI arguments, and one that runs the program. Regardless, this refactor can be done in another pull request.

FrancoGiachetta: I see. Yes, I think doing that in another PR would be better

gabrielbosio (dismissed):

gabrielbosio (approved):

gabrielbosio (approved):

FrancoGiachetta (approved):

PR #1206, opened 2025-04-30
Improve sierra2casm-dbg

JulianGCalderon:
This PR makes some improvements to the sierra2casm-dbg utility:
- Renames it to casm-data-flow.
- Adds a README with instructions on how to use it.
- Moves dependencies to workspace

Depends on #1202


debug_utils/casm-data-flow/Makefile
```diff
@@ -0,0 +1,19 @@
+usage:
+	@echo "Usage:"
+	@echo "    build:        Builds casm-data-flow."
+	@echo "    check:        Checks format and lints."
+	@echo "    clean:        Cleans the built artifacts."
```
FrancoGiachetta, line 5:
```suggestion
	@echo "build: Builds casm-data-flow."
	@echo "check: Checks format and lints."
	@echo "clean: Cleans the built artifacts."
```

FrancoGiachetta (changes requested):

debug_utils/casm-data-flow/Makefile
FrancoGiachetta, line 5:
```suggestion
usage:
	@echo "Usage:"
	@echo "build: Builds casm-data-flow."
	@echo "check: Checks format and lints."
	@echo "clean: Cleans the built artifacts."
```

FrancoGiachetta (changes requested):

debug_utils/casm-data-flow/Makefile
JulianGCalderon: I kept the leading spaces, to make it consistent with the other Makefiles in the repo. I removed some the leading spaces.

debug_utils/casm-data-flow/README.md
```diff
@@ -0,0 +1,52 @@
...
+Final Gas: 18446744073709544205
+```
+
+We will need the trace and memory files generated, to find a path from the starting gas, to the final gas.
+
+```bash
+cargo run -- --program-path programs/fibonacci_starknet.casm.json --trace-path programs/fibonacci_starknet.trace --memory-path programs/fibonacci_starknet.memory -s 18446744073709551615 -t 18446744073709544205
+```
+
+The binary prints the full path from the starting gas to the final gas. This is useful:
```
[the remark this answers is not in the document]
JulianGCalderon: Done

FrancoGiachetta (approved):

gabrielbosio (changes requested):
Looks good but what do you think about changing the crate name to `casm_data_flow` (with underscores)?
I know this would imply changing the `sierra-emu` crate in the future but now that we have this PR I think we can follow the same name style than the workspace.

gabrielbosio (approved):
> Looks good but what do you think about changing the crate name to `casm_data_flow` (with underscores)? I know this would imply changing the `sierra-emu` crate in the future but now that we have this PR I think we can follow the same name style than the workspace.

We can do this in another PR that normalizes all names

PR #1208, opened 2025-05-05
[Sierra-Emu] Fix felt-dict's overflow, circuits libfuncs and implement missing EcPoint libfunc

FrancoGiachetta:
This PR fixes some issues:

`FeltDict` libfuncs:
- There was issue while running corelib's tests which produced an overflow. This was related with the amount of initial gas passed to the executor which was `u64::MAX`. It was change to match the amount of gas needed by the test.
- `felt252_dict_get` libfunc was always returning the default value of a Felt, which was intented only in cases where the dict didn't contain the key received as argument.

`Circuits` libfuncs:
- Implements missing branch in `circuit_try_into_circuit_modulus`.
- Implements `circuit_u96_limbs_less_than_guarantee_verify` libfunc which was noop before.
- Fixes `circuit_add_input` which wasn't checking correctly when to stop adding inputs.
- Removes a panic when trying the invert a non-invertible circuit input.

"EcPoint" libfuncs
- Implements `ec_zero` libfunc

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


debug_utils/sierra-emu/Cargo.toml
```diff
@@ -23,6 +23,7 @@ cairo-lang-utils.workspace = true
 clap = { version = "4.5.26", features = ["derive"] }
 k256 = "0.13.4"
 keccak = "0.1.5"
+num-integer = "0.1.46"
```
JulianGCalderon, line 26: This dependency is defined in the workspace also. Could you point to it?

src/metadata/trace_dump.rs
```diff
@@ -572,14 +572,21 @@ pub mod trace_dump_runtime {
...
                     for i in 0..n_outputs {
-                        let size = u384_layout.pad_to_align().size();
                         let current_ptr = value_ptr.byte_add(size * i);
                         let current_value = current_ptr.as_ref();
                         values.push(BigUint::from_bytes_le(current_value));
                     }
 
-                    Value::CircuitOutputs(values)
+                    let current_ptr = value_ptr.byte_add(size * n_outputs);
+                    let modulus = BigUint::from_bytes_le(current_ptr.as_ref());
```
JulianGCalderon, line 584: The modulus is stored as a struct with four limbs. I'm not sure this conversion works

src/metadata/trace_dump.rs
```diff
@@ -572,14 +572,21 @@ pub mod trace_dump_runtime {
...
+
                     for i in 0..n_outputs {
-                        let size = u384_layout.pad_to_align().size();
                         let current_ptr = value_ptr.byte_add(size * i);
                         let current_value = current_ptr.as_ref();
                         values.push(BigUint::from_bytes_le(current_value));
                     }
 
-                    Value::CircuitOutputs(values)
+                    let current_ptr = value_ptr.byte_add(size * n_outputs);
```
JulianGCalderon, line 583: This doesn't take into account padding. I think its better to calculate the layout manually, using the methods in: https://doc.rust-lang.org/beta/std/alloc/struct.Layout.html

debug_utils/sierra-emu/tests/tests/circuits_failure.cairo
JulianGCalderon, line 1: Why did you remove this file?

debug_utils/sierra-emu/tests/common/mod.rs
```diff
@@ -20,7 +20,9 @@ pub fn value_to_felt(value: &Value) -> Vec<Felt> {
             costs.mul_mod.into(),
         ],
         Value::CircuitModulus(value) => vec![value.into()],
-        Value::Circuit(data) | Value::CircuitOutputs(data) => data.iter().map(Felt::from).collect(),
+        Value::Circuit(data) | Value::CircuitOutputs { circuits: data, .. } => {
+            data.iter().map(Felt::from).collect()
+        }
```
JulianGCalderon, line 25: The circuit outputs now contains the modulo. We should write that value also, right?

debug_utils/sierra-emu/src/vm/felt252_dict_entry.rs
```diff
@@ -36,23 +36,39 @@ pub fn eval_get(
...
+            smallvec![
+                Value::FeltDictEntry {
+                    ty,
+                    data,
+                    count,
+                    key
+                },
+                value,
+            ],
+        )
```
JulianGCalderon, line 70: The `felt252_dict_entry_get` has two return branches?

```rust
extern fn felt252_dict_entry_get<T>(
    dict: Felt252Dict<T>, key: felt252,
) -> (Felt252DictEntry<T>, T) nopanic;
```

I think it has only one.

debug_utils/sierra-emu/src/vm/felt252_dict_entry.rs
```diff
@@ -36,23 +36,39 @@ pub fn eval_get(
     };
     assert_eq!(info.ty, ty);
 
-    let default_value = Value::default_for_type(registry, &info.ty);
-    data.insert(key, default_value.clone());
+    if let std::collections::hash_map::Entry::Vacant(e) = data.entry(key) {
```
JulianGCalderon, line 39: Maybe we can use `entry(key).or_insert(default)` instead of this branching.

debug_utils/sierra-emu/src/vm/circuit.rs
```diff
@@ -12,8 +12,73 @@ use cairo_lang_sierra::{
...
+}
+
+fn find_nullifier(num: &BigUint, modulus: &BigUint) -> BigUint {
+    let ExtendedGcd { gcd, .. } = num
+        .to_bigint()
+        .unwrap()
+        .extended_gcd(&modulus.to_bigint().unwrap());
+    let gcd = gcd.to_biguint().unwrap();
+
+    // If there's no inverse, find the value which nullifys the operation
```
JulianGCalderon, line 78: What is a nullifier?

```suggestion
    // If there's no inverse, find the value which nulifies the operation
```

debug_utils/sierra-emu/src/vm/circuit.rs
```diff
@@ -48,45 +113,30 @@ pub fn eval(
...
+fn eval_add_input(
+    registry: &ProgramRegistry<CoreType, CoreLibfunc>,
+    info: &SignatureAndTypeConcreteLibfunc,
     args: Vec<Value>,
 ) -> EvalAction {
     let [Value::Circuit(mut values), Value::Struct(members)]: [Value; 2] = args.try_into().unwrap()
     else {
         panic!()
     };
-    assert_ne!(values.len(), values.capacity());
```
JulianGCalderon, line 60: I would keep this assert. It would mean that we have a bug in the implementation.

JulianGCalderon: Maybe we can change it to a debug_assert instead.

debug_utils/sierra-emu/src/vm/circuit.rs
```diff
@@ -48,45 +113,30 @@ pub fn eval(
...
 
-    let [Value::U128(l0), Value::U128(l1), Value::U128(l2), Value::U128(l3)]: [Value; 4] =
-        members.try_into().unwrap()
-    else {
-        panic!()
+    let n_inputs = match registry.get_type(&info.ty).unwrap() {
+        CoreTypeConcrete::Circuit(CircuitTypeConcrete::Circuit(info)) => info.circuit_info.n_inputs,
+        _ => panic!(),
     };
```
JulianGCalderon, line 129: Is `n_inputs` the same as `values.capacity()`.

debug_utils/sierra-emu/src/vm/circuit.rs
```diff
@@ -193,65 +256,86 @@ pub fn eval_get_output(
...
+    let [Value::Struct(garantee)]: [Value; 1] = args.try_into().unwrap() else {
+        panic!()
+    };
+    let limb_count = info.limb_count;
+    let Value::Struct(gate) = garantee.first().unwrap() else {
+        panic!();
+    };
+    let Value::Struct(modulus) = garantee.get(1).unwrap() else {
+        panic!();
+    };
```
JulianGCalderon, line 286: I think this is a better easier to read:
```suggestion
    let Value::Struct(gate) = garantee[0] else {
        panic!();
    };
    let Value::Struct(modulus) = garantee[1].unwrap() else {
        panic!();
    };
```

debug_utils/sierra-emu/src/vm/circuit.rs
```diff
@@ -193,65 +256,86 @@ pub fn eval_get_output(
...
-pub fn eval_u96_limbs_less_than_guarantee_verify(
+fn eval_u96_limbs_less_than_guarantee_verify(
     _registry: &ProgramRegistry<CoreType, CoreLibfunc>,
-    _info: &ConcreteU96LimbsLessThanGuaranteeVerifyLibfunc,
-    _args: Vec<Value>,
+    info: &ConcreteU96LimbsLessThanGuaranteeVerifyLibfunc,
+    args: Vec<Value>,
 ) -> EvalAction {
-    EvalAction::NormalBranch(0, smallvec![Value::Unit])
+    let [Value::Struct(garantee)]: [Value; 1] = args.try_into().unwrap() else {
```
JulianGCalderon, line 277:
```suggestion
    let [Value::Struct(guarantee)]: [Value; 1] = args.try_into().unwrap() else {
```

debug_utils/sierra-emu/src/vm/circuit.rs
```diff
@@ -193,65 +256,86 @@ pub fn eval_get_output(
...
+    else {
+        panic!();
+    };
+    let Value::BoundedInt {
+        value: modulus_last_limb,
+        ..
+    } = &modulus[limb_count - 1]
+    else {
+        panic!();
+    };
```
JulianGCalderon, line 300: Not really related to this pull request, but we do this everywhere (match a specific variant of a `Value`). It would be great to have a generic utility that makes the code shorter. We can take it into account and implement it in another pull request.

debug_utils/sierra-emu/src/vm/circuit.rs
```diff
@@ -260,18 +344,42 @@ pub fn eval_failure_guarantee_verify(
         panic!()
     };
 
-    EvalAction::NormalBranch(0, smallvec![rc96, mul_mod, Value::Unit])
+    let limbs_cout = match registry
```
JulianGCalderon, line 347:
```suggestion
    let limbs_count = match registry
```

debug_utils/sierra-emu/src/vm/circuit.rs
```diff
@@ -331,22 +439,27 @@ pub fn eval_try_into_circuit_modulus(
...
+    let [Value::BoundedInt { range, mut value }]: [Value; 1] = args.try_into().unwrap() else {
         panic!()
     };
     assert_eq!(range, BigInt::ZERO..(BigInt::from(1) << 96));
 
+    // offset by the lower bound to get the actual value
+    if range.start > BigInt::ZERO {
+        value = range.start;
+    }
+
```
JulianGCalderon, line 463: I think you meant to use `+=` instead of `=`. Regardless, I'm not sure this is necessary. Offsetting required in Cairo Native due to how we store bounded ints, but is it necessary in Sierra Emu?

JulianGCalderon (changes requested):

src/metadata/trace_dump.rs
FrancoGiachetta: Oh, why isn't it an `u384` like the other circuit inputs? I'll change it then

debug_utils/sierra-emu/src/vm/circuit.rs
FrancoGiachetta: You are right, it is no necessary. I'll remove it

src/metadata/trace_dump.rs
FrancoGiachetta: Why isn't it taking it into account? Isn't size already aligned?

FrancoGiachetta: Another thing which is confusing me. The layout used for u384_layout is built like this:
`let u384_layout = Layout::from_size_align(48, 16).unwrap()`
However, in types/circuit.rs, it is built like this: `get_integer_layout(384)`, which equivalent to: `Layout::from_size_align((384 as usize).next_multiple_of(8) >> 3, 16)`. Simplifying it, it would be `Layout::from_size_align(24, 16)`, which makes sense since padding it would result in 24 * 16 = 384. Am I missing something?

debug_utils/sierra-emu/tests/tests/circuits_failure.cairo
FrancoGiachetta: It was intended to panic since that was the way we handle. However that was changed in this PR. Apart from that, there's a corelib test which already tests this so there's no purpose for it anymore

debug_utils/sierra-emu/tests/common/mod.rs
FrancoGiachetta: Fixed

debug_utils/sierra-emu/src/vm/circuit.rs
FrancoGiachetta: Done

debug_utils/sierra-emu/src/vm/circuit.rs
FrancoGiachetta: Done

debug_utils/sierra-emu/src/vm/circuit.rs
FrancoGiachetta: Done

debug_utils/sierra-emu/src/vm/circuit.rs
FrancoGiachetta: I'm not following you. What would this utility do?

debug_utils/sierra-emu/Cargo.toml
FrancoGiachetta: Done

src/metadata/trace_dump.rs
JulianGCalderon: > Why isn't it taking it into account? Isn't size already aligned?

This line `value_ptr.byte_add(size * n_outputs)` asumes that there is no padding between the end of the array and the start of the next value.

Isn't `(384 as usize).next_multiple_of(8) >> 3` equal to 48?

debug_utils/sierra-emu/src/vm/circuit.rs
JulianGCalderon: Mmh, I'm not really sure. But if we find a way to remove all the boilerplate it would make the code much more cleaner. Its not really related to this PR, we can discuss it another time.

src/metadata/trace_dump.rs
FrancoGiachetta: I think not, but I may be wrong. This `(384 as usize).next_multiple_of(8)` evaluates to 384, since it is multiple of 8. And 384 >> 3 is 24, right?

JulianGCalderon: 384 >> 3 is equal to 48. Right shifting by 3 is the same as dividing by 8 (if multiple of 8)

FrancoGiachetta: Yes, you are right. I made a mistake there sorry

debug_utils/sierra-emu/src/vm/felt252_dict_entry.rs
FrancoGiachetta: branching is needed anyways since I need to increase count if the key is not present. But I will simplify it with this. Thanks!

FrancoGiachetta: Done

src/metadata/trace_dump.rs
JulianGCalderon: No problem.

debug_utils/sierra-emu/src/vm/circuit.rs
FrancoGiachetta: If the value is non-invertible a nullifier returned instead. It's a value which satisfies the equation:
`value * nullifier ≡ 0(modulus)`

debug_utils/sierra-emu/src/vm/circuit.rs
FrancoGiachetta: There's a test in the corelib which, for some reason passes to this libfunc a vec with capacity 0. This is the test: `core::test::circuit_test::test_fill_inputs_loop `. In fact, this is why i removed the `assert_ne!(values.len(), values.capacity())`. I think is preferable to compere with a constant rather than the capacity.

FrancoGiachetta: My intuition is that this is because the test seems to first create a circuit without input, and then it adds them one by one. So that could explain why the circuit gets created with capacity == 0

src/metadata/trace_dump.rs
FrancoGiachetta: Fixed.

src/metadata/trace_dump.rs
```diff
@@ -565,21 +565,46 @@ pub mod trace_dump_runtime {
...
+                        let mut gate_value = [0u8; 48];
+                        for j in 0..4 {
+                            (outputs_layout, limb_offset) =
+                                outputs_layout.extend(u96_layout).unwrap();
+                            let current_ptr = circuits_ptr.byte_add(limb_offset);
+                            let current_value = current_ptr.as_ref();
+                            gate_value[(12 * j)..(12 + 12 * j)].copy_from_slice(current_value);
+                        }
+                        values.push(BigUint::from_bytes_le(&gate_value));
+                    }
```
JulianGCalderon, line 589: The layout of the _circuit outputs_ is:
```C
struct {
  u384[n_outputs],
  u384,
}
```

That means that the `full_layout` can be built with:
```C
full_layout = layout_extend(array_layout, struct_layout)
```

Where `array_layout` is built with:
```C
array_layout = layout_repeat(struct_layout, n_outputs)
```

And `struct_layout` is built with:

```C
struct_layout = layout_repeat(u96_layout, 4)
```

Instead of calling `extend` manually after each iteration. Can you use `layout_repeat` to build both the `u384_layout` and the `array_layout`. I think something like this would work and make the implementation much cleaner.

JulianGCalderon (changes requested):

src/metadata/trace_dump.rs
JulianGCalderon: Also, for reading an u384 struct, maybe we can do something similar to what is done for U256 and create a new struct:

```
    #[repr(C, align(16))]
    pub struct U256 {
        pub lo: u128,
        pub hi: u128,
    }
```

FrancoGiachetta: Made it cleaner now. Talking about the struct proposal, I don't know if that would work. I need a struct like this one:
```
 #[repr(C, align(16))]
    pub struct U384 {
        pub limb_0: u128,
        pub limb_1: u128,
        pub limb_2: u128,
        pub limb_3: u128,
    }
```
Which would be more than 384 bits, maybe we could replace them for `[u8; 12]` but I think that would be overcomplicating things, especially because we would need to change the whole sierra-emu's circuits implementation.

debug_utils/sierra-emu/src/vm/felt252_dict_entry.rs
```diff
@@ -29,17 +29,20 @@ pub fn eval_get(
...
     assert_eq!(info.ty, ty);
 
-    let default_value = Value::default_for_type(registry, &info.ty);
-    data.insert(key, default_value.clone());
-
-    let count = count + 1;
+    let value = data
+        .entry(key)
+        .or_insert_with(|| {
+            count += 1;
```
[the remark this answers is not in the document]
FrancoGiachetta: Fixed

debug_utils/sierra-emu/src/vm/circuit.rs
```diff
@@ -12,8 +12,73 @@ use cairo_lang_sierra::{
...
+fn find_nullifier(num: &BigUint, modulus: &BigUint) -> BigUint {
+    let ExtendedGcd { gcd, .. } = num
+        .to_bigint()
+        .unwrap()
+        .extended_gcd(&modulus.to_bigint().unwrap());
+    let gcd = gcd.to_biguint().unwrap();
+
+    // If there's no inverse, find the value which nulifies the operation
+    modulus / gcd
+}
```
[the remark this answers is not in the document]
FrancoGiachetta: Done

src/metadata/trace_dump.rs
JulianGCalderon: Right, but doesn't u96 and u128 occupy the same space? At least thats what `get_integer_layout` indicates. Although it looks bad, maybe its not a good idea...

JulianGCalderon: Regarding this comment:
> especially because we would need to change the whole sierra-emu's circuits implementation.
I don't understand what you mean. Why would changing this thing in the trace dump affect the sierra emu?

FrancoGiachetta: In sierra-emu, we represent u384 as `BigUint`s. Using this struct not only would imply changing the implementation but also we would need to convert it to BigUint or to some other representation which would allow as to operate with it (like we currently do with u256).

FrancoGiachetta: Oh, nothing. I've just realized it would affect sierra emu

debug_utils/sierra-emu/src/vm/circuit.rs
```diff
@@ -12,8 +12,75 @@ use cairo_lang_sierra::{
...
+/// If the value is non-invertible a nullifier is returned instead. A nullifier
+/// is value which satisfies this following equation: num * nullifier ≡ 0(modulus).
+fn find_nullifier(num: &BigUint, modulus: &BigUint) -> BigUint {
+    let ExtendedGcd { gcd, .. } = num
+        .to_bigint()
+        .unwrap()
+        .extended_gcd(&modulus.to_bigint().unwrap());
+    let gcd = gcd.to_biguint().unwrap();
+
+    // If there's no inverse, find the value which nulifies the operation
```
JulianGCalderon, line 80: This comment can now be erased, as its already documented in the function signature.
```suggestion
```

src/metadata/trace_dump.rs
```diff
@@ -565,21 +569,41 @@ pub mod trace_dump_runtime {
...
-                        let current_ptr = value_ptr.byte_add(size * i);
-                        let current_value = current_ptr.as_ref();
-                        values.push(BigUint::from_bytes_le(current_value));
+                        let output_limbs = (0..4)
+                            .flat_map(|j| {
+                                let offset = u96_layout.size() * j + u384_struct_layout.size() * i;
+                                *value_ptr.byte_add(offset).as_ref()
+                            })
+                            .collect::<Vec<u8>>();
+                        values.push(BigUint::from_bytes_le(&output_limbs));
```
JulianGCalderon, line 593: Converting a struct with u96 limbs to a BigInt seems to be a common operation (its done 4 times in this file). Could you create an auxiliary function so that its easier to understand?

JulianGCalderon (changes requested):
Looking good, I left two small comments.

src/metadata/trace_dump.rs
JulianGCalderon: The function could receive the pointer and number of limbs, and return a BigInt?

FrancoGiachetta: Yes, I'll do it that way. Though there's an exception with `U96LimbsLessThanGuarantee`, which kind of similar but it returns each limb as a `BigUint`. So it's actually used 3 times in the file

FrancoGiachetta: Done

debug_utils/sierra-emu/src/vm/circuit.rs
FrancoGiachetta: Done

JulianGCalderon (approved):
LGTM, I suggested a small correction

JulianGCalderon (approved):
LGTM

src/metadata/trace_dump.rs
```diff
@@ -640,7 +655,40 @@ pub mod trace_dump_runtime {
...
+                                value: BigInt::from_bytes_le(Sign::Plus, current_ptr.as_ref()),
+                            }
+                        })
+                        .collect::<Vec<_>>();
+
+                    let modulus_ptr = value_ptr.byte_add(u384_struct_layout.size());
+
+                    let modulus_limbs = (0..info.limb_count)
+                        .map(|i| {
+                            let current_ptr = modulus_ptr.byte_add(u96_layout.size() * i);
```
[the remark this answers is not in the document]
FrancoGiachetta: Done

PR #1223, opened 2025-05-09
[Serra-Emu] Fix some libfuncs

FrancoGiachetta:
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


debug_utils/sierra-emu/src/vm/int.rs
```diff
@@ -242,32 +239,64 @@ fn eval_from_felt(
...
 
-    if value > range.lower && value <= range.upper {
+    let value = {
+        let value_bigint = value_felt.to_biguint();
+        if value_bigint > half_prime {
+            (prime - value_bigint).to_bigint().unwrap() * BigInt::from(-1)
+        } else {
+            value_felt.to_bigint()
+        }
+    };
```
gabrielbosio, line 256: It seems this logic treats `value_bigint` as if it where a signed integer. The function that contains this logic is also called from `eval_unsigned` (See [this link](https://github.com/lambdaclass/cairo_native/blob/8a5021f440a0840157b0248a39c89394dff4da12/debug_utils/sierra-emu/src/vm/int.rs#L120)) so I don't follow how this makes `eval_unsigned` work.

gabrielbosio (approved):
Looks good, I've left a small question

debug_utils/sierra-emu/src/vm/int.rs
FrancoGiachetta: There are two functions: `eval_signed` and `eval_unsigned`, each of which take a generic value `T`. This value T implements `IntMulTraits + SintTraits` or `IntMulTraits + IsZeroTraits + UintTraits` depending on the function. This is done so that we can get the libfuncs we need to implement for each group of ints. However, this file tries to be generic on the implementation (as we've done in native) whenever posible. This is one of those cases, `eval_from_felt`'s implementation works for signed ints as well as for unsigned ints. Due to this, the felt that this function receives may represent a negetive value, hence we check `value_bigint > half_prime` and perform the conversion in that case

FrancoGiachetta:
Question answered.

debug_utils/sierra-emu/src/vm/int.rs
gabrielbosio: Ok but is it possible that `eval_unsigned` calls `eval_from_felt` and `value_bigint > half_prime`? Is that condition still valid in that case?

JulianGCalderon (approved):
LGTM, I'd add a description before merging it.

PR #1236, opened 2025-05-19
Gather compilation statistics

JulianGCalderon:
Gather compilation statistics


JulianGCalderon (commented):
It now contains an acceptable set of statistics. We should define how would this API be exposed. Right now it always save statistics to disk when using the contract executor. It would be good to expose is to the user instead somehow.

src/statistics.rs
```diff
@@ -0,0 +1,93 @@
...
+/// It should be completely filled at the end of the compilation.
+#[derive(Default, Serialize)]
+pub struct Statistics {
+    /// Number of types defined in the sierra.
+    pub sierra_type_count: Option<usize>,
+    /// Number of libfuncs defined in the sierra.
+    pub sierra_libfunc_count: Option<usize>,
+    /// Number of statements contained in the sierra.
+    pub sierra_statement_count: Option<usize>,
+    /// Number of user functions defined in the sierra.
```
gabrielbosio, line 15: Nitpicking:

```suggestion
    /// Number of types defined in the Sierra code.
    pub sierra_type_count: Option<usize>,
    /// Number of libfuncs defined in the Sierra code.
    pub sierra_libfunc_count: Option<usize>,
    /// Number of statements contained in the Sierra code.
    pub sierra_statement_count: Option<usize>,
    /// Number of user functions defined in the Sierra code.
```

gabrielbosio (approved):
Looks good, just a small comment

gabrielbosio (approved):

src/executor/contract.rs
```diff
@@ -236,10 +269,18 @@ impl AotContractExecutor {
...
+        let object_data = crate::module_to_object(&module, opt_level, clone_option_mut!(stats))?;
+        if let Some(&mut ref mut stats) = stats {
+            stats.object_size_bytes = Some(object_data.len());
+        }
 
         // Build the shared library into the lockfile, to avoid using a tmp file.
-        crate::object_to_shared_lib(&object_data, &lock_file.0)?;
+        crate::object_to_shared_lib(&object_data, &lock_file.0, clone_option_mut!(stats))?;
+
+        let compilation_time = pre_compilation_instant.elapsed().as_millis();
```
FrancoGiachetta, line 280: I think this instant is is getting to much noise from the statistics gathering overhead and other things which happen in between. Apart from that, it seems to be included the link time right? Shouldn't we separate the compilation time and the linking?

PR #1267, opened 2025-07-01
Implement Libfunc Counter

FrancoGiachetta:
Implements the Libfunc Counter. Used to count every call of each libfunc in an execution. All libfuncs that have been compiled will be taken into account (even if they never get called, in which case the counter will be 0). An example of execution a program with this feature is as follows:
```json
{
  "struct_construct<Tuple<felt252>>": 2,
  "function_call<user@core::panic_with_const_felt252::<375233589013918064796019>>": 0,
  "function_call<user@core::panic_with_const_felt252::<2138404743268917596156934978104436>>": 0,
  "struct_construct<Tuple<Unit>>": 1,
  "store_temp<Tuple<core::panics::Panic, Array<felt252>>>": 0,
  "store_temp<RangeCheck>": 4000027,
  "const_as_immediate<Const<felt252, 375233589013918064796019>>": 0,
  "branch_align": 4000028,
  "withdraw_gas": 2000012,
  "felt252_add": 2000010,
  "disable_ap_tracking": 2000013,
  "const_as_immediate<Const<felt252, 1>>": 2000012,
  "enum_match<core::panics::PanicResult::<(core::felt252,)>>": 2,
  "enum_init<core::panics::PanicResult::<((),)>, 1>": 0,
  "const_as_immediate<Const<felt252, 10>>": 1,
  "store_temp<core::panics::PanicResult::<(core::felt252,)>>": 2,
  "const_as_immediate<Const<felt252, 0>>": 2,
  "function_call<user@core::panic_with_const_felt252::<123770447868417490207208308>>": 0,
  "store_temp<GasBuiltin>": 2000015,
  "const_as_immediate<Const<felt252, 2138404743268917596156934978104436>>": 0,
  "enum_init<core::panics::PanicResult::<(core::felt252,)>, 0>": 2,
  "felt252_is_zero": 2000014,
  "redeposit_gas": 2000013,
  "struct_construct<Unit>": 1,
  "struct_deconstruct<Tuple<felt252>>": 2,
  "dup<felt252>": 4000022,
  "function_call<user@fib_2M::fib_2M::fib>": 2000012,
  "enum_init<core::panics::PanicResult::<((),)>, 0>": 1,
  "store_temp<core::panics::PanicResult::<((),)>>": 1,
  "store_temp<felt252>": 6000039,
  "const_as_immediate<Const<felt252, 123770447868417490207208308>>": 0,
  "struct_construct<Tuple<core::panics::Panic, Array<felt252>>>": 0,
  "struct_construct<core::panics::Panic>": 0,
  "drop<felt252>": 4,
  "const_as_immediate<Const<felt252, 55>>": 1,
  "array_new<felt252>": 0,
  "enum_init<core::panics::PanicResult::<(core::felt252,)>, 1>": 0,
  "array_append<felt252>": 0,
  "function_call<user@core::panic_with_felt252>": 0,
  "felt252_sub": 2000012,
  "const_as_immediate<Const<felt252, 2000000>>": 1,
  "drop<NonZero<felt252>>": 2000010,
  "const_as_immediate<Const<felt252, 3428715265070152839999402611900589559274061538403385725837342803058381925627>>": 1
}
```

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/cast.rs
+++ b/src/libfuncs/cast.rs
@@ -55,7 +55,7 @@ pub fn build_downcast<'ctx, 'this>(
     _metadata: &mut MetadataStorage,
     info: &DowncastConcreteLibfunc,
 ) -> Result<()> {
-    let range_check = super::increment_builtin_counter(context, entry, location, entry.arg(0)?)?;
+    let range_check = entry.arg(0)?;
     let src_value: Value = entry.arg(1)?;
 
     if info.signature.param_signatures[1].ty == info.signature.branch_signatures[0].vars[1].ty {
@@ -247,6 +247,52 @@ pub fn build_downcast<'ctx, 'this>(
             }
         };
 
+        // Incrementing the range check depends on whether the source range can hold a felt252 or not
+        let range_check = if info.from_range.is_full_felt252_range() {
+            let rc_size = BigInt::from(1) << 128;
+            // If the range can contain a felt252, how the range check is increased depends on whether the value is in bounds or not:
+            // * If it is in bounds, we check whether the destination range size is less than range check size. If it is, increment
+            //   the range check builtin by 2. Otherwise, increment it by 1.
+            //   https://github.com/starkware-libs/cairo/blob/v2.12.0-dev.1/crates/cairo-lang-sierra-to-casm/src/invocations/range_reduction.rs#L87
+            // * If it is not in bounds, increment the range check builtin by 3.
+            //   https://github.com/starkware-libs/cairo/blob/v2.12.0-dev.1/crates/cairo-lang-sierra-to-casm/src/invocations/range_reduction.rs#L79
+            super::increment_builtin_counter_by_if(
+                context,
+                entry,
+                location,
+                range_check,
+                if dst_range.size() < rc_size { 2 } else { 1 },
+                3,
+                is_in_bounds,
+            )?
+        } else {
+            match (lower_check, upper_check) {
+                (Some(_), None) | (None, Some(_)) => {
+                    // If either the lower or the upper bound was checked, increment the range check builtin by 1.
+                    // * In case the lower bound was checked: https://github.com/starkware-libs/cairo/blob/v2.12.0-dev.1/crates/cairo-lang-sierra-to-casm/src/invocations/casts.rs#L135
+                    // * In case the upper bound was checked: https://github.com/starkware-libs/cairo/blob/v2.12.0-dev.1/crates/cairo-lang-sierra-to-casm/src/invocations/casts.rs#L111
+                    super::increment_builtin_counter_by(context, entry, location, range_check, 1)?
+                }
+                (Some(lower_check), Some(upper_check)) => {
+                    let is_in_range =
+                        entry.append_op_result(arith::andi(lower_check, upper_check, location))?;
+
+                    // If the result is in range, increment the range check builtin by 2. Otherwise, increment it by 1.
+                    // https://github.com/starkware-libs/cairo/blob/v2.12.0-dev.1/crates/cairo-lang-sierra-to-casm/src/invocations/casts.rs#L160
+                    super::increment_builtin_counter_by_if(
+                        context,
+                        entry,
+                        location,
+                        range_check,
+                        2,
+                        1,
+                        is_in_range,
+                    )?
+                }
+                (None, None) => range_check,
+            }
+        };
+
         let dst_value = if dst_ty.is_bounded_int(registry)? && dst_range.lower != BigInt::ZERO {
             let dst_offset = entry.const_int_from_type(
                 context,
@@ -268,6 +314,7 @@ pub fn build_downcast<'ctx, 'this>(
         } else {
             dst_value
         };
+
         helper.cond_br(
             context,
             entry,


src/bin/cairo-native-run.rs
```diff
@@ -281,6 +308,35 @@ fn main() -> anyhow::Result<()> {
...
+    if let Some(libfunc_counter_output) = args.libfunc_counter_output {
+        use std::collections::HashMap;
+
+        let counters =
+            cairo_native::metadata::libfunc_counter::libfunc_counter_runtime::LIBFUNC_COUNTER
+                .lock()
+                .unwrap();
+        assert_eq!(counters.len(), 1);
+
+        let libfunc_counter = counters.values().next().unwrap();
```
gabrielbosio, line 321: I do not fully get this part:

- Does `LIBFUNC_COUNTER` hold several counters?
- Why the `libfunc_counter` variable is set to the second value of `LIBFUNC_COUNTER`?

src/compiler.rs
```diff
@@ -640,6 +653,22 @@ fn compile_func(
                         },
                     };
 
+                    #[cfg(feature = "with-libfunc-counter")]
+                    {
+                        // Can't fail since that would we got a libfunc which is not defined in the program
+                        let libfunc_idx = libfunc_indexes.get(&invocation.libfunc_id).unwrap();
```
gabrielbosio, line 659: How can we be sure that `libfunc_indexes` has the same keys from `program.statements`?

src/metadata/libfunc_counter.rs
```diff
@@ -0,0 +1,338 @@
...
+                MemRefType::new(IntegerType::new(context, 64).into(), &[], None, None),
+                location,
+            ))
+            .unwrap();
+
+        block.append_op_result(memref::load(libfunc_counter_id_ptr, &[], location))
+    }
+
+    /// Indexes the array of counters and increments
+    /// the one relative to the given libfunc index
```
gabrielbosio, line 160: With `one` you mean `array` or `counter`?

src/metadata/libfunc_counter.rs
```diff
@@ -0,0 +1,338 @@
+#![cfg(feature = "with-libfunc-counter")]
+//! The libfunc counter feature is used to generate information counting how many time a libfunc has been called.
+//!
+//! When this feature is used, the compiler will call the important methods:
```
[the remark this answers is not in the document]
FrancoGiachetta: That they are part of the main functionality. I wanted to say "an important method". I'll change it

src/compiler.rs
FrancoGiachetta: Because of how it is constructed:
```rust
let libfunc_indexes = program
        .libfunc_declarations
        .iter()
        .enumerate()
        .map(|(idx, libf)| (libf.id.clone(), idx))
        .collect::<HashMap<ConcreteLibfuncId, usize>>();
```
This is done once, before compiling each libfunc in `src/compiler.rs:157`. Each key represents a libfunc that has been declared. If we had a key from the invocation statement that is not included in that hashmap, that that we mean that there's an error in the sierra program since we got to invoke a libfunc that has not been declared.

src/bin/cairo-native-run.rs
gabrielbosio: Ok, makes sense

src/compiler.rs
gabrielbosio: > If we had a key from the invocation statement that is not included in that hashmap, that that we mean that there's an error in the sierra program since we got to invoke a libfunc that has not been declared.

Nice, it would be great if the comment of line 658 is rephrased adding this line

src/metadata/libfunc_counter.rs
gabrielbosio, line 1: Some thoughts related to this file:

1. Why is the libfunc counter API separated in two modules instead of using only one? For example, the compiler calls `count_libfunc` from the runtime module but `build_array_counter` and `store_array_counter` are called from the root module.
2. If it is necessary to have two separated modules for the API then it would be great if the documentation includes what does each module do and (if it is possible or makes sense) the reason behind this separation in modules.
3. In line 4 I would not call the three methods listed below as `important`. If they are the only methods used by external files to count libfuncs then these 3 methods are the Libfunc Counter API.

FrancoGiachetta: 1. To have a separation from what belongs to MLIR, and what belongs to Rust (runtime).
2. Yes, I'll add docs if you agree with (1.).
3. I can change it to `main` if you prefer.

src/bin/cairo-native-run.rs
```diff
@@ -57,6 +59,11 @@ struct Args {
     /// The output path for the libfunc profilling results
     profiler_output: Option<PathBuf>,
 
+    #[cfg(feature = "with-libfunc-counter")]
+    #[arg(long)]
+    /// The output path for the execution trace
```
JulianGCalderon, line 64: Is this comment valid?

src/metadata/libfunc_counter.rs
```diff
@@ -0,0 +1,307 @@
...
+            u32_ty,
+        )?;
+
+        let value_counter = block.load(context, location, value_counter_ptr, u32_ty)?;
+        let value_incremented = block.addi(value_counter, k1, location)?;
+
+        block.store(context, location, value_counter_ptr, value_incremented)?;
+
+        Ok(())
+    }
```
JulianGCalderon, line 203: I'm not convinced that this is the right approach.

Right know, to obtain the pointer to the array, we must call an extern function `get_array_counter`. This is more complex than required. I think this feature can be achieved without relying on invoking Rust functions from the compiled code, and without relying on a global hashmap.

- Have each contract declare an MLIR global variable called `libfunc_counts` (or so). It should contain a pointer to a dinamically allocated array.
- Before executing each statement, obtain the array pointer, find the index of the associated libfunc, and increment it by 1. We should be able to obtain the array pointer by accessing the MLIR global previously defined.
- At the start of the execution, inside of `invoke_dynamic`, or `run`, alloc the array and set global's value to the arrays pointer. At the end of the execution, don't forget to replace the global's value with the previous pointer.
- To expose the libfunc counts to the user, we could return it from the `invoke_dynamic`/`run` functions.

This way, the functionality would be completely hidden from the user (as opposed to the current approach, that requires changes in `cairo-native-run`).

JulianGCalderon (changes requested):

src/metadata/libfunc_counter.rs
FrancoGiachetta: I agree It would be better the array be build and restored inside either invoke_dynamic or run. It should be possible with the current implementation. However I don't so much complexity in calling the a rust function. Despite this, it could be done if what we want is to avoid rust as much as possible. I actually did it this way because I thought it was simpler.

PR #1278, opened 2025-07-10
Fix downcast range check counting

FrancoGiachetta:
Native was always incrementing the range check by 1, this was incorrect. Whenever both upper and below checks are performed, range check must be increment by 2.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/cast.rs
+++ b/src/libfuncs/cast.rs
@@ -55,7 +55,8 @@ pub fn build_downcast<'ctx, 'this>(
     _metadata: &mut MetadataStorage,
     info: &DowncastConcreteLibfunc,
 ) -> Result<()> {
-    let range_check = super::increment_builtin_counter(context, entry, location, entry.arg(0)?)?;
+    let mut range_check: Value = entry.arg(1)?;
+
     let src_value: Value = entry.arg(1)?;
 
     if info.signature.param_signatures[1].ty == info.signature.branch_signatures[0].vars[1].ty {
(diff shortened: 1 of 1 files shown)


gabrielbosio (dismissed):

JulianGCalderon (changes requested):
I'm not sure that the behaviour is exactly like this.

- If the downcast **may** only overflow above, then we increment range check by 1: [compiler](https://github.com/starkware-libs/cairo/blob/dc8b4f0b2e189a3b107b15062895597588b78a46/crates/cairo-lang-sierra-to-casm/src/invocations/casts.rs?plain=1#L111).
- If the downcast **may** only overflow below, then we increment range check by 1: [compiler](https://github.com/starkware-libs/cairo/blob/dc8b4f0b2e189a3b107b15062895597588b78a46/crates/cairo-lang-sierra-to-casm/src/invocations/casts.rs?plain=1#L135). 
- If the downcast cannot overflow, we don't increment range check: [compiler](https://github.com/starkware-libs/cairo/blob/dc8b4f0b2e189a3b107b15062895597588b78a46/crates/cairo-lang-sierra-to-casm/src/invocations/casts.rs?plain=1#L93).
- If the downcast may overflow from below and above, then its more complex: [compiler](https://github.com/starkware-libs/cairo/blob/dc8b4f0b2e189a3b107b15062895597588b78a46/crates/cairo-lang-sierra-to-casm/src/invocations/casts.rs?plain=1#L160).
  - If the downcast is valid, we increment by 2.
  - If the downcast is not valid, we increment by 1.

At the start of the sierra to casm code, if the "from" type has the size of a full felt252 size, then the logic is different: [compiler](https://github.com/starkware-libs/cairo/blob/dc8b4f0b2e189a3b107b15062895597588b78a46/crates/cairo-lang-sierra-to-casm/src/invocations/casts.rs?plain=1#L35). I don't yet understand that function, but we should probably take it into account.

Issue #1275, opened 2025-07-10
Tracking: fix `range_check` counting


FrancoGiachetta:
## starknet.rs

- [x] `build_storage_base_address_from_felt252()`
- [x] `build_storage_address_try_from_felt252()`
- [x]  `build_class_hash_try_from_felt252()`
- [x] `build_contract_address_try_from_felt252()` 

## cast.rs

- [x] `build_downcast()`

## felt252_dict.rs

- [x] `build_squash()` 

## int_range.rs

- [x] `build_int_range_try_new()` 

## int.rs

- [x] `build_divmod()`
- [x] `build_from_felt252()`
- [x] `build_mul_guarantee_verify()`
- [x] `build_square_root()`
- [x] `build_u128s_from_felt252()`
- [x] `build_operation()`

## bounded_int.rs

- [x] `build_divrem()` 

## bytes31.rs

- [x] `build_from_felt252()` 

## mem.rs

- [x] `store_temp()`
- [x] `finalize_locals()`
- [x] `rename()`

## ec.rs

- [x] `build_point_from_x()`

## pedersen.rs

- [x] `build_pedersen()` 

## poseidon.rs

- [x] `build_hades_permutation()` 

## uint256.rs

- [x] `build_divmod()`
- [x] `build_square_root()`
- [x] `build_u256_guarantee_inv_mod_n()`

## uint512.rs

- [x] `build_divmod_u256()`

## box.rs

- [x] `ForwardSnapshot`

## circuit.rs
- [x] `build_eval()`
- [x] `build_u96_guarantee_verify()`

PR #1296, opened 2025-07-11
Fix build_divmod() range_check increment

DiegoCivi:
Libfunc: `build_divmod()`
- [Native](https://github.com/lambdaclass/cairo_native/blob/470083391c0c4a55e8860e1993389b0b029a53d1/src/libfuncs/int.rs#L298): increments range_check builtin by 1
- [Compiler](https://github.com/starkware-libs/cairo/blob/96625b57abee8aca55bdeb3ecf29f82e8cea77c3/crates/cairo-lang-sierra-to-casm/src/invocations/int/unsigned.rs#L151C1-L155C11): increments the range_check builtin depending on the **BoundedIntDivRemAlgorithm**. If its the variant **KnownSmallRhs** the range_check is incremented by 3, otherwise is incremented by 4.

**Changes**
- `lhs_range` and `rhs_range` had to be computed so we could get the `BoundedIntDivRemAlgorithm` and decide how much the increment of the builtin should be.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/int.rs
+++ b/src/libfuncs/int.rs
@@ -1,6 +1,6 @@
 use super::{BlockExt, LibfuncHelper};
 use crate::{
-    error::Result,
+    error::{panic::ToNativeAssertError, Result},
     metadata::MetadataStorage,
     native_panic,
     types::TypeBuilder,
@@ -8,6 +8,7 @@ use crate::{
 };
 use cairo_lang_sierra::{
     extensions::{
+        bounded_int::BoundedIntDivRemAlgorithm,
         core::{CoreLibfunc, CoreType, CoreTypeConcrete},
         int::{
             signed::{SintConcrete, SintTraits},
@@ -19,6 +20,7 @@ use cairo_lang_sierra::{
         },
         is_zero::IsZeroTraits,
         lib_func::SignatureOnlyConcreteLibfunc,
+        ConcreteLibfunc,
     },
     program_registry::ProgramRegistry,
 };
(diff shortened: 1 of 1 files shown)


src/libfuncs/int.rs
```diff
@@ -310,18 +312,40 @@ fn build_diff<'ctx, 'this>(
...
+    let lhs_ty = registry.get_type(&info.param_signatures()[1].ty)?;
+    let rhs_ty = registry.get_type(&info.param_signatures()[2].ty)?;
+    let lhs_range = lhs_ty.integer_range(registry)?;
+    let rhs_range = rhs_ty.integer_range(registry)?;
+
+    let div_rem_algorithm = BoundedIntDivRemAlgorithm::try_new(&lhs_range, &rhs_range)
+        .to_native_assert_error(&format!(
+            "div_rem of ranges: lhs = {:#?} and rhs= {:#?} is not supported yet",
+            &lhs_range, &rhs_range
+        ))?;
```
FrancoGiachetta, line 335: This can be simplified using the `?`.

src/libfuncs/int.rs
```diff
@@ -1253,15 +1293,26 @@ mod test {
...
                 None,
             )?;
 
-            assert_eq!(result.builtin_stats.range_check, 1);
+            match div_rem_algorithm {
+                BoundedIntDivRemAlgorithm::KnownSmallRhs => {
+                    assert_eq!(result.builtin_stats.range_check, 3)
+                }
+                BoundedIntDivRemAlgorithm::KnownSmallQuotient { q_upper_bound: _ }
+                | BoundedIntDivRemAlgorithm::KnownSmallLhs { lhs_upper_sqrt: _ } => {
```
FrancoGiachetta, line 1305: You could simplify this expression like this
```suggestion
                BoundedIntDivRemAlgorithm::KnownSmallQuotient { .. }
                | BoundedIntDivRemAlgorithm::KnownSmallLhs { .. } => {
```

src/libfuncs/int.rs
```diff
@@ -1245,6 +1274,17 @@ mod test {
...
+        // Get the range to create the BoundedIntDivRemAlgorithm
+        let range = Range {
+            lower: T::min_value().into(),
+            upper: T::max_value().into() + BigInt::one(),
+        };
+        let div_rem_algorithm = BoundedIntDivRemAlgorithm::try_new(&range, &range)
+            .to_native_assert_error(&format!(
+                "div_rem of ranges: lhs = {:#?} and rhs= {:#?} is not supported yet",
+                &range, &range
+            ))?;
```
FrancoGiachetta, line 1286: This would wl

src/libfuncs/int.rs
DiegoCivi: `.try_new()` returns an `Option` so it won't let me use the `?`. Should we use `.ok_or()`? In that case, any suggestions for what type of error to return?

src/libfuncs/int.rs
DiegoCivi: Done in [`7163a9c`](https://github.com/lambdaclass/cairo_native/pull/1296/commits/7163a9c0af142574c2799b8a2e9b784da1064a93)

src/libfuncs/int.rs
```diff
@@ -1253,15 +1293,26 @@ mod test {
 
             let result = executor.invoke_dynamic(
                 &program.funcs[0].id,
-                &[perm[0].into(), perm[1].into()],
+                &[Value::from(perm[0]), Value::from(perm[1])],
```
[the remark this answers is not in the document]
DiegoCivi: Yes, when I added the `Into<BigInt>` restriction to `T` it started to complain with:

> mismatched types
expected `Value`, found `BigInt`

DiegoCivi:
Solves part of #1275

PR #1297, opened 2025-07-11
fix range check counting in `libfuncs/mem.rs`

FrancoGiachetta:
Fixes range check counting in every libfunc contained in `libfuncs/mem.rs`.

The following libfuncs use `build_noop()` which adds to the builtin if it is a parameter. So for example if we call `store_temp<RangeCheck>` it will add 1 to the range check builtin. However, I dont see that the same is being done on the compilers side: 
    - `store_temp()`
    - `finalize_locals()`
    - `rename()`

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


JulianGCalderon (approved):
LGTM!

DiegoCivi:
Solves part of #1275

PR #1300, opened 2025-07-11
Fix build_u128s_from_felt252() range_check increment

DiegoCivi:
Libfunc: `build_u128s_from_felt252()`
- [Native](https://github.com/lambdaclass/cairo_native/blob/470083391c0c4a55e8860e1993389b0b029a53d1/src/libfuncs/int.rs#L818): increments range_check builtin by 1
- [Compiler](https://github.com/starkware-libs/cairo/blob/96625b57abee8aca55bdeb3ecf29f82e8cea77c3/crates/cairo-lang-sierra-to-casm/src/invocations/int/unsigned128.rs#L234): if *is_u128* is not equal to 0 range_check is incremented by 1. Else, if it is equal to 0, the range_check is incremented by 3.

**Changes**
- Use increment_builtin_counter_by_if() to add the corresponding amount to the range_check builtin according to the condition

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/int.rs
+++ b/src/libfuncs/int.rs
@@ -907,6 +907,7 @@ mod test {
         fmt::Display,
         mem,
         ops::{BitAnd, BitOr, BitXor},
+        u128,
     };
 
     fn test_bitwise<T>() -> Result<(), Box<dyn std::error::Error>>
@@ -1808,7 +1809,11 @@ mod test {
             let lo = u128::from_le_bytes(value_bytes[..16].try_into().unwrap());
             let hi = u128::from_le_bytes(value_bytes[16..].try_into().unwrap());
 
-            assert_eq!(result.builtin_stats.range_check, 1);
+            if value < Felt::from(BigInt::from(u128::MAX + 1)) {
+                assert_eq!(result.builtin_stats.range_check, 3);
+            } else {
+                assert_eq!(result.builtin_stats.range_check, 1);
+            }
             assert_eq!(
                 result.return_value,
                 Value::Enum {
--- a/src/libfuncs/int.rs
+++ b/src/libfuncs/int.rs
@@ -835,13 +835,16 @@ fn build_u128s_from_felt252<'ctx, 'this>(
     let k0 = entry.const_int_from_type(context, location, 0, target_ty)?;
     let is_wide = entry.cmpi(context, CmpiPredicate::Ne, hi, k0, location)?;
 
+    // The sierra-to-casm compiler uses the range check builtin a total of 3 times if the
+    // is_wide condition is true. Otherwise it will be used once.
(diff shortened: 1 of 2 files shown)


src/libfuncs/int.rs
```diff
@@ -837,6 +835,16 @@ fn build_u128s_from_felt252<'ctx, 'this>(
...
+    let range_check = super::increment_builtin_counter_by_if(
+        context,
+        entry,
+        location,
+        entry.arg(0)?,
+        1,
+        3,
+        is_wide,
+    )?;
+
```
JulianGCalderon, line 847: Shouldn't this be inverted?

The original number is separated into low and high bits. `is_wide` is true if the high bits is different from zero (thus, the original number is greather than 128 bits)

If the `is_wide` is false, then we are in the "not overflow" case, we should increment range check by only 1. If its true, we should increment it by 3.

JulianGCalderon (changes requested):
I'm not sure the logic is correct. Let's recheck this just in case.

src/libfuncs/int.rs
DiegoCivi: You are right, both things done in [`e440139`](https://github.com/lambdaclass/cairo_native/pull/1300/commits/e44013923450f1159df4a761744775993e55689f). Thanks!

FrancoGiachetta (approved):
LGTM

src/libfuncs/int.rs
```diff
@@ -837,6 +835,19 @@ fn build_u128s_from_felt252<'ctx, 'this>(
     let k0 = entry.const_int_from_type(context, location, 0, target_ty)?;
     let is_wide = entry.cmpi(context, CmpiPredicate::Ne, hi, k0, location)?;
 
+    // The sierra-to-casm compiler uses the range check builtin a total of 3 times if the
+    // is_wide condition is true. Otherwise it will be used once.
+    // https://github.com/starkware-libs/cairo/blob/96625b57abee8aca55bdeb3ecf29f82e8cea77c3/crates/cairo-lang-sierra-to-casm/src/invocations/int/unsigned128.rs#L234
```
gabrielbosio, line 840: Using `is_wide` while talking about sierra-to-casm compiler may be confusing since in the compiler context there is no `is_wide` variable.

gabrielbosio (changes requested):
Left a small comment about the documentation

src/libfuncs/int.rs
JulianGCalderon: True, I'd rephrase it as "when the value is greater than u128 max".

DiegoCivi: True! Done in [`2ab800d`](https://github.com/lambdaclass/cairo_native/pull/1300/commits/2ab800d86f1b8ef668dae44f260fdea754e0eb4e)

JulianGCalderon (approved):
LGTM!

gabrielbosio (approved):

DiegoCivi:
Solves part of #1275

PR #1301, opened 2025-07-11
Compare builtins in rust and corelib tests

### LINKED ISSUE #1289: Add builtin comparison to Cairo Native tests
Some Cairo Native tests already compare execution with Cairo VM to ensure that the output and gas usage is the same. We should also add support for comparing builtin stats.

This should straightforward, as the comparing logic is already encapsulated in `compare_outputs`: 

https://github.com/lambdaclass/cairo_native/blob/470083391c0c4a55e8860e1993389b0b029a53d1/tests/common.rs?plain=1#L474-L483

### LINKED ISSUE #1290: Compare builtin usage in Corelib test.
We already have a way for executing the corelib tests with Cairo Native.

It would be great if we could compare the test execution with the Cairo VM, and verify that the final builtin stats are correct.

Currently, we only execute the tests once (only with Cairo Native). This change would imply executing each test twice (once with Cairo VM, and once with Cairo Native).

JulianGCalderon:
Adds support for comparing builtin usage in both Rust and Cairo tests.

Closes #1289
Closes #1290

We used to have a hack for ignoring the compilation of certain tests. This worked for running with Cairo Native, but fails when trying to run tests with Cairo VM.

I could not make the hack work for Cairo VM, so I decided to remove this feature, and instead removing the corelib tests manually by uploading the modified corelib to the repository. Its not the best solution, but its the only way I could find to make it work.


src/bin/utils/test.rs
```diff
@@ -200,15 +243,84 @@ pub fn run_tests(
...
+                        native_builtins.insert("mul_mod", native_result.builtin_stats.mul_mod);
+
+                        for builtin_name in vm_builtins.keys() {
+                            native_builtins.entry(builtin_name).or_default();
+                        }
+                        native_builtins
+                    };
+
+                    for (builtin_name, native_counter) in native_builtins {
+                        let vm_counter = vm_builtins.get(builtin_name).cloned().unwrap_or_default();
```
FrancoGiachetta, line 307: I think this would be more of an error, right? There's a possibility that `native_counter` would also be the default value and pass the  `native_counter != vm_counter`  check.

FrancoGiachetta (commented):
Some comment

FrancoGiachetta (approved):

PR #1302, opened 2025-07-11
Fix build_from_felt252() range_check increment

DiegoCivi:
Libfunc: `build_from_felt252()`
- [Native](https://github.com/lambdaclass/cairo_native/blob/470083391c0c4a55e8860e1993389b0b029a53d1/src/libfuncs/int.rs#L338): increments range_check builtin by 1
- [Compiler](https://github.com/starkware-libs/cairo/blob/96625b57abee8aca55bdeb3ecf29f82e8cea77c3/crates/cairo-lang-sierra-to-casm/src/invocations/range_reduction.rs#L26): if *in_range* is equal to 0 adds the range_check 3 times because it falls into validate_under_limit(). If it is not equal to 0 it jumps to **InRange** which adds the range_check once. Then if `out_range.size() < rc_size` the range_check is added again (only in the case that the jump to **InRange** is done)

Co-authored-by: @FrancoGiachetta 

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/int.rs
+++ b/src/libfuncs/int.rs
@@ -6,7 +6,6 @@ use crate::{
     types::TypeBuilder,
     utils::{ProgramRegistryExt, PRIME},
 };
-use ark_ff::One;
 use cairo_lang_sierra::{
     extensions::{
         core::{CoreLibfunc, CoreType, CoreTypeConcrete},
@@ -453,42 +452,16 @@ fn build_from_felt252<'ctx, 'this>(
     //   the range is greater than or equal to the size of the range check.
     // With the range check size being 2**128
     // https://github.com/starkware-libs/cairo/blob/96625b57abee8aca55bdeb3ecf29f82e8cea77c3/crates/cairo-lang-sierra-to-casm/src/invocations/range_reduction.rs#L26
-    let range_check = entry.append_op_result(scf::r#if(
-        is_in_range,
-        &[IntegerType::new(context, 64).into()],
-        {
-            let region = Region::new();
-            let block = region.append_block(Block::new(&[]));
-            // When shifting left 128 times we would need 256 bits since
-            // it would not fit in 128. To avoid this we substract one from
-            // the two elements we want to compare.
-            let rc_size_value = (BigInt::from(1) << 128) - BigInt::one();
-            let rc_size = block.const_int(context, location, rc_size_value, 128)?;
-            let out_range = block.const_int(context, location, threshold_size - 1, 128)?;
-            let condition =
-                block.cmpi(context, CmpiPredicate::Ult, out_range, rc_size, location)?;
-            let range_check = super::increment_builtin_counter_by_if(
-                context,
-                &block,
-                location,
-                entry.arg(0)?,
-                2,
-                1,
-                condition,
-            )?;
-            block.append_operation(scf::r#yield(&[range_check], location));
-            region
-        },
-        {
-            let region = Region::new();
-            let block = region.append_block(Block::new(&[]));
-            let range_check =
-                super::increment_builtin_counter_by(context, &block, location, entry.arg(0)?, 3)?;
-            block.append_operation(scf::r#yield(&[range_check], location));
-            region
-        },
+    let rc_size = BigInt::from(1) << 128;
+    let range_check = super::increment_builtin_counter_by_if(
+        context,
+        entry,
         location,
-    ))?;
+        entry.arg(0)?,
+        if threshold_size < rc_size { 2 } else { 1 },
+        3,
+        is_in_range,
+    )?;
 
     let value = entry.trunci(value, value_ty, location)?;
 
--- a/src/libfuncs/int.rs
+++ b/src/libfuncs/int.rs
@@ -446,9 +446,11 @@ fn build_from_felt252<'ctx, 'this>(
     };
 
     // The sierra-to-casm compiler uses the range check builtin a total of:
-    //      - 3 times if is_in_range is false
-    //      - 2 times if is_in_range is true and out_range < rc_size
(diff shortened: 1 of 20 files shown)


src/libfuncs/int.rs
```diff
@@ -445,6 +445,50 @@ fn build_from_felt252<'ctx, 'this>(
...
+                context,
+                &block,
+                location,
+                entry.arg(0)?,
+                2,
+                1,
+                condition,
+            )?;
+            block.append_operation(scf::r#yield(&[range_check], location));
+            region
```
JulianGCalderon, line 479: This logic can be resolved in compile time, there is no need to encode this condition into MLIR.

You can do something like (pseudocode):
```rust
let range_check = if range < rc_size {
    increment_builtin_counter_by(2)
} else {
    increment_builtin_counter_by(1)
};
```

src/libfuncs/int.rs
```diff
@@ -1386,16 +1431,43 @@ mod test {
...
+                        );
+                    }
+                }
+                Err(_) => assert_eq!(
+                    result.builtin_stats.range_check, 3,
+                    "Type: {}  Lower: {}  Upper: {}  Value: {}",
+                    type_id, lower_bound, upper_bound, value
+                ),
+            }
+
```
JulianGCalderon, line 1464: Doing `T::try_from` might not be necessary.

The variable `target` already contains the expected value:
- if `target == None`, the value is out of range
- if `target == Some(_)`, the value is in range

Please verify that this is the case, as I might be wrong. If so, we can remove the `T::try_from` and also remove the added traits.

src/libfuncs/int.rs
JulianGCalderon: Also, this _might_ be a good use case for using the `select` operation, instead of the `if` operation. It might be worth to look into it, as using the `if` operation adds much more code than a simpler `select`.

JulianGCalderon (changes requested):
Hey @DiegoCivi, looks good. I made some suggestion on how to improve it.

src/libfuncs/int.rs
DiegoCivi: Mmm maybe I am getting confused, but is it really possible with select? The thing is that we have 3 possible cases here:
- If `is_in_range` is false the _range_check_ is used 3 times.
- If `is_in_range` is true:
    - If `range < rc_size` then the _range_check_ is used 2 times.
    - If `range >= rc_size` then the _range_check_ is used once.
 
So I need to first check the `is_in_range` and then depending on its value I might need to check another condition.

FrancoGiachetta: I think we could do something like: 

```rust 
select(is_in_range,                  //  fst condition
    if range < rc_size {
        increment_builtin_counter_by(2)
    } else {
        increment_builtin_counter_by(1)
    },                              // true case
    increment_builtin_counter_by(3) // false case
)
```
Which should be simpler enough. Didn't realize of this when we were implementing it.

src/libfuncs/int.rs
```diff
@@ -445,6 +445,50 @@ fn build_from_felt252<'ctx, 'this>(
         (is_in_range, value)
     };
 
+    // The sierra-to-casm compiler uses the range check builtin a total of:
+    //      - 3 times if the value is not within the range.
+    //      - 2 times if the value is within the range and the size of
+    //        the range is less than the size of the range check.
+    //      - 1 time if the value is within the range and the size of
+    //        the range is greater than or equal to the size of the range check
```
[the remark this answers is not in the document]
DiegoCivi: Done in [`94328fc`](https://github.com/lambdaclass/cairo_native/pull/1302/commits/94328fc696118c018bde6f1d128331484f1ac956)

src/libfuncs/int.rs
JulianGCalderon: When converting felts to singed integers, the first half of the range is for positive values, and the second half of the range is for negative values.

For negative values (where `felt > half_prime`), the actual value is calculate as `-(PRIME - felt)`. So if `felt = Felt::MAX` (that is `PRIME - 1`), then the integer value would be `-(PRIME - (PRIME - 1))`, which is equal to `-1`.

src/libfuncs/int.rs
```diff
@@ -1470,13 +1488,38 @@ mod test {
         for (value, target) in data {
             let result = executor.invoke_dynamic(&program.funcs[0].id, &[value.into()], None)?;
 
-            assert_eq!(result.builtin_stats.range_check, 1);
+            // If try_from(value) fails, means the value is out of range for T
```
JulianGCalderon, line 1491: This comment is outdated, right?

src/libfuncs/int.rs
```diff
@@ -1470,13 +1488,38 @@ mod test {
...
+            match target {
+                Some(_) => {
+                    let range_size = T::max_value().into() - T::min_value().into() + BigInt::one();
+                    let rc_size = BigInt::from(1) << 128;
+                    if range_size < rc_size {
+                        assert_eq!(
+                            result.builtin_stats.range_check, 2,
+                            "Type: {}  Value: {}",
```
JulianGCalderon, line 1499: To keep consistency:
```suggestion
                            "Difference in range_check count. Type: {}  Value: {}",
```

src/libfuncs/int.rs
DiegoCivi: My bad. Done in [`57c0368`](https://github.com/lambdaclass/cairo_native/pull/1302/commits/57c0368a83856cdf49262407d63982ad5a57e2e1)

gabrielbosio (approved):

JulianGCalderon (approved):
LGTM!

DiegoCivi:
Solves part of #1275

PR #1303, opened 2025-07-11
Fix int_operation range_check counting

FrancoGiachetta:
Fixes range check counting in libfunc `int_operation` for every int type.

- `build_operation()` 
    - unsigned 
        - [Native](https://github.com/lambdaclass/cairo_native/blob/470083391c0c4a55e8860e1993389b0b029a53d1/src/libfuncs/int.rs#L520): adds range_check 1 time
        - [Compiler](https://github.com/starkware-libs/cairo/blob/96625b57abee8aca55bdeb3ecf29f82e8cea77c3/crates/cairo-lang-sierra-to-casm/src/invocations/int/unsigned.rs#L142C13-L145C11): increments range_check by 1
    - signed 
        - Native: same as above
        - [Compiler](https://github.com/starkware-libs/cairo/blob/96625b57abee8aca55bdeb3ecf29f82e8cea77c3/crates/cairo-lang-sierra-to-casm/src/invocations/int/signed.rs#L146C9-L148C10): if the jump to **IsInRange** is NOT done, then the range_check is only incremented by 1. If the jump is made, then the range_check first is added once and then if `!(min_value == i128::MIN && max_value == i128::MAX)` it will be incremented an additional time
    - unsigned128 
        - Native: same as above
        - [Compiler](https://github.com/starkware-libs/cairo/blob/96625b57abee8aca55bdeb3ecf29f82e8cea77c3/crates/cairo-lang-sierra-to-casm/src/invocations/int/unsigned128.rs#L22C10-L25C11): increments range_check by 1
    - signed128 ❌
        - Native: same as above
        - [Compiler](https://github.com/starkware-libs/cairo/blob/96625b57abee8aca55bdeb3ecf29f82e8cea77c3/crates/cairo-lang-sierra-to-casm/src/invocations/int/signed128.rs#L19C15-L21C10): same case as with the *signed*

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/int.rs
+++ b/src/libfuncs/int.rs
@@ -1,6 +1,7 @@
 use super::{BlockExt, LibfuncHelper};
 use crate::{
     error::Result,
+    libfuncs::{increment_builtin_counter, increment_builtin_counter_by_if},
     metadata::MetadataStorage,
     native_panic,
     types::TypeBuilder,
@@ -526,10 +527,10 @@ fn build_operation<'ctx, 'this>(
     metadata: &mut MetadataStorage,
     info: &IntOperationConcreteLibfunc,
 ) -> Result<()> {
-    let range_check = super::increment_builtin_counter(context, entry, location, entry.arg(0)?)?;
-
+    let mut range_check = increment_builtin_counter(context, entry, location, entry.arg(0)?)?;
     let value_ty = registry.get_type(&info.signature.param_signatures[1].ty)?;
-    let is_signed = !value_ty.integer_range(registry)?.lower.is_zero();
+    let value_range = value_ty.integer_range(registry)?;
+    let is_signed = !value_range.lower.is_zero();
     let value_ty = value_ty.build(
         context,
         helper,
@@ -565,6 +566,31 @@ fn build_operation<'ctx, 'this>(
     )?;
 
     if is_signed {
+        let is_in_range = {
+            let k1_neg = entry.const_int_from_type(context, location, -1, value_ty)?;
+            let min_value_range =
+                entry.const_int_from_type(context, location, value_range.clone().lower, value_ty)?;
+            let min_value_range_neg = entry.muli(min_value_range, k1_neg, location)?;
+            let canonical_value = entry.addi(result, min_value_range_neg, location)?;
+            let range_size =
+                entry.const_int_from_type(context, location, value_range.size(), value_ty)?;
+
+            entry.cmpi(
+                context,
+                CmpiPredicate::Slt,
+                canonical_value,
+                range_size,
+                location,
+            )?
+        };
+        let is_not_i128 =
+            !(value_range.clone().lower == i128::MIN.into() && value_range.upper == i128::MAX.into());
+        let is_not_i128_value =
+            entry.const_int_from_type(context, location, is_not_i128, value_ty)?;
+        let is_in_range_and_not_i128 = entry.append_op_result(arith::andi(is_not_i128_value, is_in_range, location))?;
+
+        range_check = increment_builtin_counter_by_if(context, entry, location, range_check, 1,0, is_in_range_and_not_i128)?;
+
         let block_in_range = helper.append_block(Block::new(&[]));
         let block_overflow = helper.append_block(Block::new(&[]));
 
--- a/src/libfuncs.rs
+++ b/src/libfuncs.rs
@@ -497,6 +497,31 @@ fn increment_builtin_counter_by<'ctx: 'a, 'a>(
     ))
 }
 
+fn increment_builtin_counter_by_if<'ctx: 'a, 'a>(
+    context: &'ctx Context,
+    block: &'ctx Block<'ctx>,
+    location: Location<'ctx>,
+    value_to_inc: Value<'ctx, '_>,
+    true_amount: impl Into<BigInt>,
+    false_amount: impl Into<BigInt>,
+    condition: Value<'ctx, '_>,
+) -> crate::error::Result<Value<'ctx, 'a>> {
+    let true_amount_value = block.const_int(context, location, true_amount, 64)?;
+    let false_amount_value = block.const_int(context, location, false_amount, 64)?;
+
+    let true_incremented =
+        block.append_op_result(arith::addi(value_to_inc, true_amount_value, location))?;
+    let false_incremented =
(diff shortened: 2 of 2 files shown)


src/libfuncs/int.rs
```diff
@@ -579,6 +588,58 @@ fn build_operation<'ctx, 'this>(
...
+                canonical_value_minus_one,
+                range_size_minus_one,
+                location,
+            )?
+        };
+        let is_not_i128 = !(value_range.clone().lower == i128::MIN.into()
+            && value_range.upper == i128::MAX.into());
+        let is_not_i128_value = entry.const_int(context, location, is_not_i128, 1)?;
+        let is_in_range_and_not_i128 =
+            entry.append_op_result(arith::andi(is_not_i128_value, is_in_range, location))?;
```
JulianGCalderon, line 629: Can we reuse the logic that we already have for checking if the value is in range? We can move the increment to the block in line 643 (`block_in_range`).

Also, as we can check if the range is i128 at compile time, there would be no need to use `increment_builtin_counter_by_if`.

src/libfuncs/int.rs
```diff
@@ -592,7 +601,24 @@ fn build_operation<'ctx, 'this>(
...
+            // https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra-to-casm/src/invocations/int/signed.rs#L105
+            if is_i128 {
+                range_check = increment_builtin_counter_by(
+                    context,
+                    block_in_range,
+                    location,
+                    range_check,
+                    1,
+                )?;
+            }
```
JulianGCalderon, line 618: Shouldn't this condition be inverted?

```rust
    if !(min_value == i128::MIN && max_value == i128::MAX) {
```

JulianGCalderon (changes requested):

src/libfuncs/int.rs
FrancoGiachetta: Yes, my bad

src/libfuncs/int.rs
```diff
@@ -540,10 +541,18 @@ fn build_operation<'ctx, 'this>(
     metadata: &mut MetadataStorage,
     info: &IntOperationConcreteLibfunc,
 ) -> Result<()> {
-    let range_check = super::increment_builtin_counter(context, entry, location, entry.arg(0)?)?;
-
+    // Increment the range check builtin by 1
```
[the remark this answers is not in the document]
FrancoGiachetta: Done

src/libfuncs/int.rs
```diff
@@ -540,10 +541,18 @@ fn build_operation<'ctx, 'this>(
...
+    // for signed ints: https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra-to-casm/src/invocations/int/signed.rs#L68
+    // for unsinged ints:
+    //     for overflowing add: https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra-to-casm/src/invocations/int/unsigned.rs#L19
+    //     for overflowing sub: https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra-to-casm/src/invocations/int/mod.rs#L67
+    // for unsigned128:
+    //     for overflowing add: https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra-to-casm/src/invocations/int/unsigned128.rs#L45
+    //.    for overflowing sub: https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra-to-casm/src/invocations/int/mod.rs#L146
```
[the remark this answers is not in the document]
FrancoGiachetta: Done

src/libfuncs/int.rs
```diff
@@ -540,10 +541,18 @@ fn build_operation<'ctx, 'this>(
...
+    let mut range_check = increment_builtin_counter(context, entry, location, entry.arg(0)?)?;
```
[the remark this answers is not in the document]
JulianGCalderon: We should use `let`, its safer.

If we mutate the variable, we *could* be returning an MLIR value that is declared in another MLIR block. I don't know how MLIR handles this situations, but its better to be safe.

For example, the following line would mutate the variable even if we are not in the block `block_in_range`.

```rust
range_check = increment_builtin_counter_by(
    context,
    block_in_range,
    location,
    range_check,
    1,
)?;
```

By using `let` we make sure that this never happens.

JulianGCalderon (approved):
LGTM!

src/libfuncs/int.rs
```diff
@@ -566,10 +567,19 @@ fn build_operation<'ctx, 'this>(
...
-
+    // Regardless of the operation range, the range check builtin pointer is always increased at least once.
+    // * for signed ints: https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra-to-casm/src/invocations/int/signed.rs#L68
+    // * for signed128: behaves the same as the signed ints case.
+    // * for unsinged ints:
+    //    * for overflowing add: https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra-to-casm/src/invocations/int/unsigned.rs#L19
+    //    * for overflowing sub: https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra-to-casm/src/invocations/int/mod.rs#L67
+    // * for unsigned128:
+    //    * for overflowing add: https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra-to-casm/src/invocations/int/unsigned128.rs#L45
+    //    * for overflowing sub: https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-sierra-to-casm/src/invocations/int/mod.rs#L146
```
gabrielbosio, line 578: I have noticed that some files have changed in the latest `cairo-lang-sierra-to-casm` versions. For example: https://github.com/starkware-libs/cairo/blob/v2.12.0-dev.1/crates/cairo-lang-sierra-to-casm/src/invocations/int/mod.rs

I think we should use a link that points to the version of `cairo-lang-sierra-to-casm` we are using in the VM. We can go further and point to a commit instead of a branch.
Also, if we are going to proceed with my suggestion, we also should review the links we added to the other libfuncs.

FrancoGiachetta: Done

gabrielbosio (approved):

PR #1317, opened 2025-07-16
Fix is_not_i128 condition for overflowing operation libfunc

JulianGCalderon:
The upper bound is exclusive, so the i128 range would be represented as `{ lower: i128::min, upper: i128::max + 1}`. 

Similarly, to check if a range is the same as the i128 range, we must check that `lower == i128::min`, and `upper - 1 == i128::max`.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/int.rs
+++ b/src/libfuncs/int.rs
@@ -647,8 +647,8 @@ fn build_operation<'ctx, 'this>(
         ));
 
         {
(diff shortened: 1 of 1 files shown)


DiegoCivi (approved):
Looks good! Nice catch

PR #1335, opened 2025-07-29
Fix walk_mlir_operations with manual iterator

JulianGCalderon:
We used to call `mlir_sys::mlirOperationWalk` directly, which is provided by the MLIR library. This works great if we don't need to access operation attributes (for example, if we are just counting the number of operations).

For some reason, when trying to access the operation, segmentation fault occurs. I don't know what causes this. To fix it, is implemented the iterator manually, instead of relying on the already implemented function.

This approach also has the benefit of being able to use a closure instead of an extern function. So regardless of this bug, I think this new approach is better.


DiegoCivi (approved):
Looks good!

PR #1340, opened 2025-07-31
Add to cairo-native-compile support for compiling sierra directly

JulianGCalderon:
Adapt the `cairo-native-compile` binary to support compiling sierra directly, instead of receiving the Cairo file/project.

It can be used with Scarb to compile Cairo projects. Scarb generates Sierra json files, which can then be compiled with this tool.

In addition to this, this PRs adds some improvements to the binary:
- Better documentation.
- Specifying default values in clap directly.
- Support for saving compilation statistics.

# Example

To compile a scarb project with Native, run:

```sh
scarb build
cairo-native-compile target/dev/a.sierra.json --sierra-json
```

### DIFF
--- a/src/bin/cairo-native-compile.rs
+++ b/src/bin/cairo-native-compile.rs
@@ -31,10 +31,10 @@ struct Args {
     /// By default, it is intrepreted as a Cairo project.
     path: PathBuf,
     /// Whether path is a single Cairo file.
-    #[arg(short, long)]
+    #[arg(short, long, group = "input")]
     single_file: bool,
     /// Whether path is a single Sierra JSON file.
-    #[arg(long)]
+    #[arg(long, group = "input")]
     sierra_json: bool,
     /// Optimization level (Valid: 0, 1, 2, 3).
     /// Values higher than 3 are considered as 3.


src/bin/cairo-native-compile.rs
```diff
@@ -1,41 +1,55 @@
...
     single_file: bool,
-    /// Allows the compilation to succeed with warnings.
+    /// Whether path is a single Sierra JSON file.
     #[arg(long)]
-    allow_warnings: bool,
-    /// Replaces sierra ids with human-readable ones.
-    #[arg(short, long, default_value_t = false)]
-    replace_ids: bool,
-    /// Optimization level, Valid: 0, 1, 2, 3. Values higher than 3 are considered as 3.
+    sierra_json: bool,
```
FrancoGiachetta, line 38: This looks mutually exclusive with `single_file`. This could cause an error if we have set both flags and we pass it a cairo program instead of a json, right?

FrancoGiachetta (commented):
Looks good overall, just a small comment. However I'm curious of something, isn't the json generated by scarb compatible with the starknet-native-compile? Regardless of the answer, I think it would be more apriopiate to implement this feature there, since `cairo-native-compile` is supposed to work with non-contract programs as far as I'm concerned.

src/bin/cairo-native-compile.rs
JulianGCalderon: Yes, they are mutually exclusive. I'll make some changes to force this behaviour.

gabrielbosio (approved):
Looks good, just a question

PR #1344, opened 2025-08-01
[Next] Use dynamic array while evaluating circuit

JulianGCalderon:
Right know, we keep a vector of MLIR values while evaluating a circuit (one value per gate). This implies that we are using a lot of MLIR values at the same time, which puts a lot of pressure on the LLVM optimization phases.

This PR changes it to use a dynamic array of gate values, which uses a single register (a pointer to the array).

# Benchmark

I compiled several circuit-heavy contract classes, and compared the compilation time before and this PR:

| Class Hash | Before | After |
| ---------- | ------ | ----- |
| 0x1b5fbe104c033025dbb7fb37011781cc9344e881b4828cdaa023a80fecafde4 | 77024 ms | 91510 ms | 
| 0x3100defca27214e5f78f25e48a5b05e45899c6834cb4d34f48384c18e14dff7 | 74235 ms | 88951 ms | 
| 0x79666cdb4fc3cbcafbd74f4ea4e2855bf455c5a7c70915f5679325c54032771 | 87430 ms | 88287 ms | 
| 0xa40ecaa08eaef629d433b15f6b017461df8bea9b4e7299eb8ba3632e32b5a5 | 96162 ms | 97812 ms | 
| 0x4ffeed293927cd56686a9038a10026a2d3b9602f789d1f163c1c4ac9a822a82 | 107289 ms | 129285 ms | 
| 0x2269858a40ea0535cb373b0c981c91b907466edf6d65bcaf669760bbee0ae4d | 98239 ms | 122565 ms | 
| 0x5ff378cb2f16804539ecb92e84f273aafbab57d450530e9fe8e87771705a673 | 88499 ms | 100990 ms | 
| 0x4edde37ca59d9dff8f4ac8945b1c4860b606abd61d74727904ad7494fccdfa9 | 88920 ms | 101077 ms | 

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/libfuncs/circuit.rs
+++ b/src/libfuncs/circuit.rs
@@ -327,7 +327,7 @@ fn build_eval<'ctx, 'this>(
         circuit_info.add_offsets.len() * ADD_MOD_BUILTIN_SIZE,
     )?;
 
-    let ([ok_block, err_block], gates) = build_gate_evaluation(
+    let [ok_block, err_block] = build_gate_evaluation(
         context,
         entry,
         location,
@@ -339,17 +339,6 @@ fn build_eval<'ctx, 'this>(
 
     // Ok case
     {
-        // We drop circuit_data, as its consumed by this libfunc.
-        if let Some(drop_overrides_meta) = metadata.get::<DropOverridesMeta>() {
-            drop_overrides_meta.invoke_override(
-                context,
-                ok_block,
-                location,
-                &info.signature.param_signatures[3].ty,
-                circuit_data,
-            )?;
-        }
-
         // Increase the mul mod builtin pointer by the number of evaluated gates.
         // If the evaluation succedes, then we assume that every gate was evaluated.
         // https://github.com/starkware-libs/cairo/blob/v2.12.0-dev.1/crates/cairo-lang-sierra-to-casm/src/invocations/circuit.rs?plain=1#L261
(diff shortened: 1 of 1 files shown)


src/types/circuit.rs
```diff
@@ -278,12 +278,13 @@ pub fn build_circuit_data<'ctx>(
 ///
 /// ## Layout:
 ///
-/// Holds N_VALUES elements, where each element is a u384 struct,
-/// A u384 struct contains 4 limbs, each a u96 integer.
+/// Holds 1 + N_VALUES + N_INPUTS elements, where each element is an u384 integer (u384i),
```
FrancoGiachetta, line 281: What does `u384i` mean?

FrancoGiachetta (commented):
Looks good, just small comment. 

This would definitely require another PR but since the use of dynamic arrays are increasing, I think it would be nice to wrap its operations inside functions. For example, to update an element, we need to offset the pointer with a `gep` operation. Then we need to load the value and operate it. And, finally, we need to store it.
Perhaps having utility functions for these arrays would be better.

src/libfuncs/circuit.rs
```diff
@@ -478,44 +437,128 @@ fn build_gate_evaluation<'ctx, 'this>(
...
+
+    let ptr_type = llvm::r#type::pointer(context, 0);
+    let u384_type = IntegerType::new(context, 384).into();
+    let u384_layout = &get_integer_layout(384);
+
+    // Alloc memory for the full circuit.
+    let circuit_length = block.const_int(
+        context,
+        location,
+        (1 + circuit_info.n_inputs + circuit_info.values.len()) * u384_layout.pad_to_align().size(),
```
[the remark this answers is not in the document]
JulianGCalderon: It is equivalent. If you prefer I can change it to use `repeat_layout` instead.

src/types/circuit.rs
JulianGCalderon: Its a shorthand for u384 integer, and helps to differentiate from u384 struct (u384s).

src/libfuncs/circuit.rs
```diff
@@ -478,44 +437,128 @@ fn build_gate_evaluation<'ctx, 'this>(
...
+        context,
+        location,
+        circuit_ptr,
+        &[GepIndex::Const(1)],
+        u384_type,
+    )?;
+    let circuit_input_length = block.const_int(
+        context,
+        location,
+        circuit_info.n_inputs * u384_layout.pad_to_align().size(),
```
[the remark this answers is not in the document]
JulianGCalderon: Done 4293c491e5fd4443f14808680892f688f55b4905.

src/libfuncs/circuit.rs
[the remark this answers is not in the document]
JulianGCalderon: Done 4293c491e5fd4443f14808680892f688f55b4905.

src/libfuncs/circuit.rs
```diff
@@ -478,44 +437,129 @@ fn build_gate_evaluation<'ctx, 'this>(
     circuit_info: &circuit::CircuitInfo,
     circuit_data: Value<'ctx, 'ctx>,
     circuit_modulus: Value<'ctx, 'ctx>,
-) -> Result<([&'this Block<'ctx>; 2], Vec<Value<'ctx, 'ctx>>)> {
-    // Throughout the evaluation of the circuit we maintain an array of known gate values
-    // Initially, it only contains the inputs of the circuit.
-    // Unknown values are represented as None
+) -> Result<[&'this Block<'ctx>; 2]> {
+    // Throughout the evaluation of the circuit we maintain an MLIR dynamic
+    // array of gate values. It will contain 1 + N_INPUTS + N_GATES elements.
```
DiegoCivi, line 442: Why does it need to have that length? I saw that the first element is just a 1 but as far as I could understand we never use that

DiegoCivi (approved):
Looks good! One minor question.

src/libfuncs/circuit.rs
```diff
@@ -464,12 +416,19 @@ fn build_eval<'ctx, 'this>(
...
+/// - The success block receives:
+///   - The evaluated circuit pointer.
+/// - The error block receives:
+///   - The evaluated circuit pointer.
+///   - The index of the first gate that could not be computed.
+///
+/// This function consumes the `circuit_data` input variable, and frees the underlying memory.
+///
+/// The original Cairo hint evaluates all gates, even in case of failure.
+/// This implementation exits on first error, as there is no need for the partial outputs yet.
```
gabrielbosio, line 431: Why there is no need for the partial outputs yet?

gabrielbosio (approved):
Looks good, just one question

FrancoGiachetta (dismissed):

gabrielbosio:
Both `Hyperfine` and `Compare Dumps` are not running since get triggered on a PR to `main` only.

PR #1350, opened 2025-08-05
Add script to compile and generete debug files

FrancoGiachetta:
This PR adds a script to compile a contract class and generate dumps for mlir, llvmir, sierra and casm. It also adds a new binary contract-to-sierra, to extract the sierra from a contract class.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


debug_utils/contract-utils/Cargo.toml
JulianGCalderon, line 1: I would add the new binary to `debug_utils` directly. I don't find any benefit on having an extra level of indirection.

scripts/compile-sierra-contract-class.sh
```diff
@@ -0,0 +1,66 @@
...
+CASM_PATH=$CLASS_HASH.casm
+OS="$(uname -s)"
+
+if [[ $OS == "Darwin" ]]; then
+    LLVM_PATH="$(brew --prefix llvm@19)"
+elif [[ $OS == "Linux" ]]; then
+    LLVM_PATH="/usr/lib/llvm-19"
+else
+    echo "Unsupported OS: $OS"
+fi
```
JulianGCalderon, line 41: This would fail if the user has LLVM installed in another directory. Maybe we should reuse the environment variables: `LLVM_SYS_191_PREFIX`

scripts/compile-sierra-contract-class.sh
```diff
@@ -0,0 +1,66 @@
...
+set -e
+
+print_usage() {
+cat <<EOF 
+Missing parameters to provide.
+
+Usage: $0 <contract_class_path> <opt_lvl> <output_path>
+
+Compiles a Sierra Contract Class, generates the MLIR and LLVMIR files 
+and saves then along with the Sierra and CASM files in <output_path>-files/.
```
JulianGCalderon, line 11: Appending something to what the user inputs seems odd. Wouldn't it be better to save the files to `<output_directory>/` directly?

Being `<output_directorty>` a script argument.

scripts/compile-sierra-contract-class.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Extract the sierra from the contract class.
+cargo run -p contract-utils --bin contract-to-sierra $CONTRACT_PATH >> $SIERRA_PATH
+
+# Lower sierra to casm
+./cairo2/bin/sierra-compile $SIERRA_PATH $CASM_PATH
+
+echo "Compiling contract class..."
+# Set NATIVE_DEBUG_DUMP to generate mlir files.
+NATIVE_DEBUG_DUMP=true cargo run --release --bin starknet-native-compile -- -O $OPT_LVL $CONTRACT_PATH $OUTPUT_PATH
```
JulianGCalderon, line 51: Unrelated to this PR, it would be good if `starknet-native-compile` had the `--mlir` option for generating the MLIR files without requiring that env var.

scripts/compile-sierra-contract-class.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
+set -e
+
+print_usage() {
+cat <<EOF 
+Missing parameters to provide.
```
JulianGCalderon, line 6: This sentence sounds a bit odd. I would put something like "expected X arguments, received Y".

scripts/compile-sierra-contract-class.sh
FrancoGiachetta: Done in: [1bfb7d2](https://github.com/lambdaclass/cairo_native/pull/1350/commits/1bfb7d20bb04882925bea7bb6377033495a52252)

scripts/compile-sierra-contract-class.sh
FrancoGiachetta: I agree, would be more straight forward.

scripts/compile-sierra-contract-class.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
+set -e
+
+print_usage() {
+cat <<EOF 
+Expected 3 arguments, received $1.
```
JulianGCalderon, line 6: This prints only the first argument. To print the number of arguments you should use:

```suggestion
Expected 3 arguments, received $#.
```

FrancoGiachetta27: The number of arguments is passed as an argument to the function. So actually `$1`, refers to that argument. If you prefer, I can use `$#` directly.

FrancoGiachetta: Actually, `$1` refers to the number of arguments (it's passed as an argument to the function). I can use `$#` directly if you prefer

JulianGCalderon: Oh sorry, you are right. You can dismiss this comment.

JulianGCalderon (dismissed):
LGTM

scripts/compile-sierra-contract-class.sh
```diff
@@ -0,0 +1,65 @@
...
+CONTRACT_PATH=$1
+OPT_LVL=$2 
+DEST_DIR=$3
+CONTRACT_CLASS_FILE=$(basename $CONTRACT_PATH)
+CLASS_HASH=${CONTRACT_CLASS_FILE%.*}
+SIERRA_PATH=$CLASS_HASH.sierra
+CASM_PATH=$CLASS_HASH.casm
+
+# Extract the sierra from the contract class.
+cargo run -p debug_utils --bin contract-to-sierra $CONTRACT_PATH >> $SIERRA_PATH
```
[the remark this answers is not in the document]
DiegoCivi: Nevermind, I just realized you move the file to a directory and this would replace the old file. You can dismiss the comment :)

DiegoCivi (approved):

scripts/compile-sierra-contract-class.sh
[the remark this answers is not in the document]
FrancoGiachetta: Done in: f4239921fc4e6208b989647ae5432ff4c4596b0b

scripts/compile-sierra-contract-class.sh
JulianGCalderon, line 1: Could you give execution permission to this file? That way it can be executed with `./scripts/compile-sierra-contrac-class.sh`

JulianGCalderon (approved):
LGTM!

PR #1357, opened 2025-08-11
MLIR implementation of euclidean algorithm

DiegoCivi:
The euclidean algorithm was being built for every inverse gate in a circuit. The aim of this PR is to improve compilation times in circuit related contracts by reducing the MLIR size and this is done by declaring the euclidean algorithm only once and the just calling it.

## Benchmark
The following are the compilation times it took with the `no_inline` attribute and with an O2 optimization level. The `Before` column contains the compilation times for the main branch ([ref](https://github.com/lambdaclass/cairo_native/commit/21cc66c9ad56a2d45889d7d9fa6f04b812a30068) -> without the optimization done in #1348), while the `After` column contains the compilations times for this branch ([ref](https://github.com/lambdaclass/cairo_native/pull/1357/commits/b2a4a6626e87a2319143e4321949550b013ac0cb) -> again, without the optimization done in #1348 ). These were ran on a MacBook Air M1 with 16 GB.

| Class Hash | Before (ms) | After (ms) | Improvement |
|------------|-------------|-----------|---------------|
|`0x1b5fbe104c033025dbb7fb37011781cc9344e881b4828cdaa023a80fecafde4` | 130730 | 110260 | 1.19x|
|`0x3100defca27214e5f78f25e48a5b05e45899c6834cb4d34f48384c18e14dff7` | 125509 | 107107 | 1.17x|
|`0x79666cdb4fc3cbcafbd74f4ea4e2855bf455c5a7c70915f5679325c54032771` | 123684 | 112580 | 1.10x|
|`0xa40ecaa08eaef629d433b15f6b017461df8bea9b4e7299eb8ba3632e32b5a5` | 132966 | 118622 | 1.12x|
|`0x4ffeed293927cd56686a9038a10026a2d3b9602f789d1f163c1c4ac9a822a82` | 177028 | 155397 | 1.14x|
|`0x2269858a40ea0535cb373b0c981c91b907466edf6d65bcaf669760bbee0ae4d` | 169722 | 149323 | 1.14x|
|`0x5ff378cb2f16804539ecb92e84f273aafbab57d450530e9fe8e87771705a673` | 138254 | 120414 | 1.15x|
|`0x4edde37ca59d9dff8f4ac8945b1c4860b606abd61d74727904ad7494fccdfa9` | 138312 | 120037 | 1.15x|

## Correctness
Blocks 1560000 to 1580000 were executed and no diffs were found when comparing with the VM

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/libfuncs/circuit.rs
```diff
@@ -475,11 +477,13 @@ fn build_eval<'ctx, 'this>(
 ///
 /// The original Cairo hint evaluates all gates, even in case of failure.
 /// This implementation exits on first error, as there is no need for the partial outputs yet.
+#[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
FrancoGiachetta: I think it's more readable to have them as separated parameters instead of using a tuple. it adds observability.

src/metadata/runtime_bindings.rs
```diff
@@ -41,6 +45,7 @@ enum RuntimeBinding {
     DictDup,
     GetCostsBuiltin,
     DebugPrint,
+    EvalCircuit,
```
FrancoGiachetta, line 48: I wouldn't name it `EvalCIrcuit`. `eval_circuit` is a libfunc to eval an entire circuit. In this PR we are moving the euclidean algorithm to the runtime. Maybe we can name it `CircuitDivMod` or something related to add, then we should mention in the implementation that we are using the euclidean algorithm.

src/metadata/runtime_bindings.rs
```diff
@@ -107,6 +113,7 @@ impl RuntimeBinding {
             RuntimeBinding::GetCostsBuiltin => {
                 crate::runtime::cairo_native__get_costs_builtin as *const ()
             }
+            RuntimeBinding::EvalCircuit => unimplemented!(),
```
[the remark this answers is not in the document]
FrancoGiachetta: No only than. Returning a null pointer could lead to undefined behaviour, I think it's betterto return `unreachable!()` since this code should never be reached.

src/metadata/runtime_bindings.rs
```diff
@@ -167,6 +174,46 @@ impl RuntimeBindingsMeta {
         )?)
     }
 
+    pub fn libfunc_build_eval<'c, 'a>(
```
FrancoGiachetta, line 177: I'd change this name to match the comment about the `EvalCIrcuit` variant.

src/metadata/runtime_bindings.rs
```diff
@@ -167,6 +174,46 @@ impl RuntimeBindingsMeta {
...
+        module: &Module,
+        block: &'a Block<'c>,
+        location: Location<'c>,
+        rhs_value: Value<'c, '_>,
+        circuit_modulus: Value<'c, '_>,
+    ) -> Result<Value<'c, 'a>>
+    where
+        'c: 'a,
+    {
+        let integer_type: Type = IntegerType::new(context, 384 * 2).into();
```
FrancoGiachetta, line 189: Why is it `384 * 2`? Shouldn't it be an `u384`? We are doing divmod operation so it should never be bigger than a `u384`, right?

src/metadata/runtime_bindings.rs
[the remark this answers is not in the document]
DiegoCivi: Awesome, didn't know about that method. Thanks! Done in [`d3567e5`](https://github.com/lambdaclass/cairo_native/pull/1357/commits/d3567e5a973638501cd898dfb9873d39c0a43759)

src/libfuncs/circuit.rs
```diff
@@ -638,17 +642,34 @@ fn build_gate_evaluation<'ctx, 'this>(
...
+                            helper.module,
+                            block,
+                            location,
+                            rhs_value,
+                            circuit_modulus,
+                        )?,
+                        None => native_panic!(
+                            "Unable to get the RuntimeBindingsMeta from MetadataStorage"
+                        ),
+                    };
```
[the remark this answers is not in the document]
DiegoCivi: Awesome, didn't know about that method. Thanks! Done in [`d96f8fd`](https://github.com/lambdaclass/cairo_native/pull/1357/commits/d96f8fd81f0ac3202ab3f136dee347ca7ef39cf4)

src/metadata/runtime_bindings.rs
DiegoCivi: Done!

src/metadata/runtime_bindings.rs
DiegoCivi: Done!

src/metadata/runtime_bindings.rs
FrancoGiachetta: That's quite better! I forgot we where calculating the inverse.

JulianGCalderon:
> `After` column contains the compilations times for this branch.

Does it include the optimization done in #1384?

src/metadata/runtime_bindings.rs
```diff
@@ -107,6 +113,7 @@ impl RuntimeBinding {
             RuntimeBinding::GetCostsBuiltin => {
                 crate::runtime::cairo_native__get_costs_builtin as *const ()
             }
+            RuntimeBinding::CircuitDivMod => unreachable!(),
```
JulianGCalderon, line 116: What if we change the `function_ptr` function to return an Option instead?

src/metadata/runtime_bindings.rs
```diff
@@ -686,3 +740,112 @@ pub fn setup_runtime(find_symbol_ptr: impl Fn(&str) -> Option<*mut c_void>) {
         }
     }
 }
+
+fn register_euclidean_algorithm_mlir_func<'ctx>(
```
JulianGCalderon, line 744: A few things about this signature:
- I would use `build` instead of `register`, as we are actually building the function declaration (and implementation).
- The algorithm is actually the _extended_ euclidean algorithm

Given this, I would rename this function to either:
- `build_extended_euclidean_algorithm_function` or, if too long,
- `build_egcd_function` with a comment stating the full name.

src/metadata/runtime_bindings.rs
```diff
@@ -686,3 +740,112 @@ pub fn setup_runtime(find_symbol_ptr: impl Fn(&str) -> Option<*mut c_void>) {
...
+    module: &Module,
+    context: &'ctx Context,
+    location: Location<'ctx>,
+    integer_type: Type<'_>,
```
JulianGCalderon, line 748: Why do we need to receive the integer_type? Can we create it inside of the function instead?

src/metadata/runtime_bindings.rs
```diff
@@ -686,3 +740,112 @@ pub fn setup_runtime(find_symbol_ptr: impl Fn(&str) -> Option<*mut c_void>) {
...
+    func_symbol: &str,
+) -> Result<()> {
+    let region = Region::new();
+
+    let entry_block = region.append_block(Block::new(&[
+        (integer_type, location),
+        (integer_type, location),
+    ]));
+
+    // The algorithm egcd works by calculating a series of remainders, each the remainder of dividing the previous two
```
JulianGCalderon, line 758:
```suggestion
    // The egcd algorithm works by calculating a series of remainders, each the remainder of dividing the previous two
```

src/metadata/runtime_bindings.rs
```diff
@@ -686,3 +740,112 @@ pub fn setup_runtime(find_symbol_ptr: impl Fn(&str) -> Option<*mut c_void>) {
...
+    // For the initial setup, r0 = b, r1 = a
+    // This order is chosen because if we reverse them, then the first iteration will just swap them
+    let remainder = entry_block.arg(0)?;
+    let prev_remainder = entry_block.arg(1)?;
+
+    // Similarly we'll calculate another series which starts 0,1,... and from which we will retrieve the modular inverse of a
```
JulianGCalderon, line 764: Could you split this comment into two lines?

src/metadata/runtime_bindings.rs
```diff
@@ -686,3 +740,112 @@ pub fn setup_runtime(find_symbol_ptr: impl Fn(&str) -> Option<*mut c_void>) {
...
+        TypeAttribute::new(llvm::r#type::function(
+            llvm::r#type::r#struct(context, &[integer_type, integer_type], false),
+            &[integer_type, integer_type],
+            false,
+        )),
+        region,
+        &[(
+            Identifier::new(context, "no_inline"),
+            Attribute::unit(context),
+        )],
```
JulianGCalderon, line 847: Add a comment explaining why is this attribute required.

src/metadata/runtime_bindings.rs
JulianGCalderon, line 744: We should add a comment explaining what the algorithm does. What are the input and output variables?

src/metadata/runtime_bindings.rs
```diff
@@ -167,6 +174,53 @@ impl RuntimeBindingsMeta {
...
+                OperationBuilder::new("llvm.call", location)
+                    .add_attributes(&[
+                        (
+                            Identifier::new(context, "callee"),
+                            FlatSymbolRefAttribute::new(context, func_symbol).into(),
+                        ),
+                        (
+                            Identifier::new(context, "no_inline"),
+                            Attribute::unit(context),
+                        ),
```
JulianGCalderon, line 214: We could remove this attribute. It should suffice to have it in the function declaration. 

https://llvm.org/docs/LangRef.html#function-attributes

> noinline: This attribute indicates that the inliner should never inline this function in any situation. This attribute may not be used together with the alwaysinline attribute.

src/metadata/runtime_bindings.rs
JulianGCalderon: Idem for the RuntimeBindingsMeta method.

src/metadata/runtime_bindings.rs
```diff
@@ -65,6 +70,7 @@ impl RuntimeBinding {
             RuntimeBinding::DictDrop => "cairo_native__dict_drop",
             RuntimeBinding::DictDup => "cairo_native__dict_dup",
             RuntimeBinding::GetCostsBuiltin => "cairo_native__get_costs_builtin",
+            RuntimeBinding::CircuitDivMod => "cairo_native__div_mod_operation",
```
JulianGCalderon, line 73: Can we rename this variant to something like ExtendedEuclideanAlgorithm or EGCD?

src/metadata/runtime_bindings.rs
JulianGCalderon: You can take the documentation from the original function. Also, we should document the RuntimeBindingsMeta method.

JulianGCalderon (changes requested):
Nice work! Left some comment to improve on.

src/metadata/runtime_bindings.rs
DiegoCivi: Outside of that function, that same type is used and thought this way I was 'building' less MLIR code. But it doesn't make too much sense. Changed in [`fa2ebca`](https://github.com/lambdaclass/cairo_native/pull/1357/commits/fa2ebca33beb5e47fb733166c8afce863fcbe666)

src/metadata/runtime_bindings.rs
DiegoCivi: Done in [`9aaddf6`](https://github.com/lambdaclass/cairo_native/pull/1357/commits/9aaddf6dab44bdd0b151e19eae1bfcb82303a432)

src/metadata/runtime_bindings.rs
DiegoCivi: Done in [`e851925`](https://github.com/lambdaclass/cairo_native/pull/1357/commits/e851925eb287e53205948bb4c2a302cb21b9f4b6)

src/metadata/runtime_bindings.rs
JulianGCalderon: Yes, in `setup_runtime` we can just skip the `None` cases.

Also, please add some documentation to `function_ptr` explaining why it can return `None` in some cases. The short explanation is that the `function_ptr` is used to setup external bindings (implemented in Rust), while some bindings are internal (implemented in MLIR, thus there is no function_ptr).

src/metadata/runtime_bindings.rs
```diff
@@ -167,6 +176,44 @@ impl RuntimeBindingsMeta {
...
 
+    /// Build if necessary the extended euclidean algorithm used in circuit inverse gates
+    pub fn extended_euclidean_algorithm<'c, 'a>(
+        &mut self,
+        context: &'c Context,
+        module: &Module,
+        block: &'a Block<'c>,
+        location: Location<'c>,
+        rhs_value: Value<'c, '_>,
+        circuit_modulus: Value<'c, '_>,
```
JulianGCalderon, line 187: Given that this is a generic implementation of egcd, we could rename arguments to `a` and `b`.  (no need to keep the circuit-related names)

```suggestion
        rhs_value: Value<'c, '_>,
        circuit_modulus: Value<'c, '_>,
```

DiegoCivi:
> Does it include the optimization done in #1384?

Nope, i'll change the description to clarify that

src/metadata/runtime_bindings.rs
DiegoCivi: Perfect, thanks for the explanation! Done in [`81fd737`](https://github.com/lambdaclass/cairo_native/pull/1357/commits/81fd73743f7769580d0c0992d05bd357b8b54d4d)

src/metadata/runtime_bindings.rs
DiegoCivi: Done in [`1065576`](https://github.com/lambdaclass/cairo_native/pull/1357/commits/10655760b362adf07e3b5ec7e3aad81f03889cec)

JulianGCalderon (approved):
Looks good!

FrancoGiachetta (approved):

gabrielbosio (approved):

PR #1400, opened 2025-09-15
Optimize circuit Compilation

FrancoGiachetta:
# Optimize circuit Compilation.

This PR changes how we handle arithmetic operations with circuits. Currently, for every gate, we extend all the operands to avoid overflows (to either u385 o u768), we generate the corresponding `arith` operations and then trunc the result back to u384. 
In x86, arithmetic operations need to go through a legalization pass. This is because x86 allows up to 2 registers per instruction, while addition, subtraction and multiplication operations need 3 registers. A circuit with enough gates can make compilation quite longer due to this. 
To avoid having all these operations inlined, this PR creates function which takes care of performing the arithmetic and return the result ready to be used. This allow to significantly reduce the compilation time since the go from thousands of `arith` operations to less than 10, and very little work for the legalization pass to perform. 
However, this comes with the downside of adding an extra indirection. Now every circuit operation reduces to almost one function call, with the exception of the inversion operation which was left as it is now since it already had an extra indirection. This generates a regression in execution time.

## Benchmarks

In both cases, compilation and execution, classes were compiled using `--opt-level 2`. Some classes may have no execution benchmarks, this is because there were no transactions invoking them.

| class | compilation BASE (s) | compilation HEAD (s) | improvement | execution BASE (ms) | execution HEAD (ms) | improvement |
|------|----------------------|-----------------------|----------|----------------------|----------------------|----------|
| 0x05ff378cb2f16804539ecb92e84f273aafbab57d450530e9fe8e87771705a673 | 340.94 | 67.23 | 5.07 | 296.76 | 291.56 | 1.01 |
| 0x4ffeed293927cd56686a9038a10026a2d3b9602f789d1f163c1c4ac9a822a82 | 448.50 | 67.19 | 6.67 | 43.37 | 46.63 | 0.93 | 
| 0x4edde37ca59d9dff8f4ac8945b1c4860b606abd61d74727904ad7494fccdfa9 | 328.67 | 67.31 | 4.88 | - | - | - |
| 0x5862777a13917417ef3bb87ebc28b071753c316b1a3fb31f9e937d73e0aa188 | 480.27 | 88.51 | 5.42 | - | - | - |

No.

## Checklist

- [ ] Linked to Github Issue.
- [ ] Unit tests added.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1996,7 +1996,7 @@ dependencies = [
  "aho-corasick",
  "bstr",
  "log",
- "regex-automata 0.4.9",
+ "regex-automata 0.4.13",
  "regex-syntax 0.8.5",
 ]
 
@@ -2116,7 +2116,7 @@ dependencies = [
  "globset",
  "log",
  "memchr",
- "regex-automata 0.4.9",
+ "regex-automata 0.4.13",
  "same-file",
  "walkdir",
  "winapi-util",
@@ -2319,7 +2319,7 @@ version = "0.22.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b5baa5e9ff84f1aefd264e6869907646538a52147a755d494517a8007fb48733"
 dependencies = [
- "regex-automata 0.4.9",
+ "regex-automata 0.4.13",
  "rustversion",
 ]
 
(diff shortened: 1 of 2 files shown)


src/metadata/runtime_bindings.rs
```diff
@@ -131,6 +134,14 @@ impl RuntimeBinding {
     }
 }
 
+// This enum is used when performing circuit arith operations.
+// Inversion is not included because it is handled separately.
+pub enum CircuitOpType {
+    Add,
+    Sub,
+    Mul,
+}
```
JulianGCalderon, line 143: Would it make sense to handle inversion here?

FrancoGiachetta: The thing with inversion is that it already has it operation-heavy part wrapped into a function. So I didn't want to add an extra indirection. Besides, that operation might fail. So the new function should also handle that too, which would make it more complex than it currently is.

PR #1431, opened 2025-10-20
Avoid using generic-array >0.14.7

FrancoGiachetta:
# Avoid using generic-array >0.14.7

Closes #NA

Recently, `generic-array` 0.14.x has been deprecated in 0.14.8. Due to this, sierra-emu, which uses `generic-array` started to fail with clippy. This is actually not an direct dependency but indirect one, from [k256](https://crates.io/crates/k256). Since we don't expect to bump it to 1.x and the deprecation is definitive (as it's been told in this [discussion](https://github.com/fizyk20/generic-array/issues/158) suggests), this PR pins `generic-array` to accept versions such that v <= 0.14.7. 

No.

## Checklist

- [ ] Linked to Github Issue.
- [ ] Unit tests added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,6 @@ cairo-native.path = "."
 clap = "4.5.23"
 colored = "2.1.0"
 criterion = "0.5.1"
-generic-array = "<=0.14.7"
 itertools = "0.14.0"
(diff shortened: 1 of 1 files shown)


JulianGCalderon (approved):
LGTM!

PR #1446, opened 2025-11-06
Implement Squashed Felt252 Dict Entries libfunc

### LINKED ISSUE #1151: [2.11.1] Implement Squashed Felt252 Dict Entries Libfunc
Cairo version 2.11.1 introduces a new libfunc that we should implement:

```cairo
extern fn squashed_felt252_dict_entries<T>(
    dict: SquashedFelt252Dict<T>,
) -> Array<(felt252, T, T)> nopanic;
```

See casm lowering for more information: https://github.com/starkware-libs/cairo/blob/1a531143af0066d3d9141d63dfbfba2e293ff8f0/crates/cairo-lang-sierra-to-casm/src/invocations/squashed_felt252_dict.rs#L6-L14

Our implementation will differ from the casm lowering because the memory layout is different.

DiegoCivi:
# Implement Squashed Felt252 Dict Entries libfunc

Implements the SquashedFelt252Dict libfunc:

- [x] SquashedFelt252DictConcreteLibfunc::IntoEntries

Closes #1151 

No.

## Checklist

- [x] Linked to Github Issue.
- [x] Unit tests added.


src/runtime.rs
```diff
@@ -296,6 +299,110 @@ pub unsafe extern "C" fn cairo_native__dict_get(
...
+/// necessary offset.
+///
+/// # Caveats
+///
+/// Each tuple has the form (felt252, T, T) = (key, first_value, last_value). 'last_value' is represents
+/// the value of the element in the dictionary and 'first_value' is always 0.
+pub unsafe extern "C" fn cairo_native__dict_into_entries(
+    dict_ptr: *const FeltDict,
+    data_prefix_offset: u64,
+    tuple_stride: u64,
```
DiegoCivi: I am receiving the size of the tuple from the MLIR code because I was unable to create it on the runtime side. I tried to do it like this:
```rust
let tuple_stride = Layout::new::<[u8; 32]>()
    .extend(dict.layout).unwrap().0
    .extend(dict.layout)
```
To my understanding, this is the way to calculate the `Layout` of a `struct<felt252, T, T>`

src/libfuncs/squashed_dict.rs
```diff
@@ -0,0 +1,330 @@
...
+    )?;
+    // Create the pointer and alloc the necessary memory
+    let ptr_ty = llvm::r#type::pointer(context, 0);
+    let nullptr = entry.append_op_result(llvm::zero(ptr_ty, location))?;
+    let array_ptr = entry.append_op_result(ReallocBindingsMeta::realloc(
+        context,
+        nullptr,
+        realloc_len,
+        location,
+    )?)?;
```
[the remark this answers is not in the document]
FrancoGiachetta: Perhaps it was overwritten by accident, but I'm still seeing the realloc.

src/runtime.rs
```diff
@@ -296,6 +300,110 @@ pub unsafe extern "C" fn cairo_native__dict_get(
...
+    let data_prefix_offset = calc_data_prefix_offset(tuple_layout);
+
+    let arr = create_mlir_array(dict, data_prefix_offset, tuple_layout.pad_to_align().size());
+
+    *array_ptr = arr;
+
+    // This libfunc consumes the dictionary, so we need to drop it.
+    // But since we are moving the elements to the array, we set the
+    // drop_fn to None so it doesn't try to apply that function on undefined memory.
+    dict.drop_fn = None;
```
FrancoGiachetta: Are we sure this is okay? Without `drop_fn` the dict won't be able to get dropped, right?

DiegoCivi: That `drop_fn` field contains the implementation of the `drop()` for the elements of the dictionary that are in the `elements` field. In `create_mlir_array()` I am moving those elements into the newly created array an putting "garbage" in the `elements` section. If I don't set that to `None`, then when the dictionary is dropped it will try to apply that `drop_fn` to the garbage that was inserted. By setting it to `None` it just deallocs the 'element' field without caring for the content. I might be wrong about this, but as of my understanding it should be able to drop the dict.

src/libfuncs/squashed_dict.rs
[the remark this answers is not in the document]
DiegoCivi: Mmm, weird. I don't see the realloc anymore, I see the alloca 🤔

src/runtime.rs
FrancoGiachetta: Ah, I see. you are dropping it manually during `cairo_native__dict_into_entries`. You can dismiss the comment.

src/libfuncs/squashed_dict.rs
```diff
@@ -0,0 +1,297 @@
...
+    // Runtime function that creates the array with its content
+    metadata
+        .get_mut::<RuntimeBindingsMeta>()
+        .ok_or(Error::MissingMetadata)?
+        .dict_into_entries(context, helper, entry, dict_ptr, array_ptr, location)?;
+
+    // Extract the array from the pointer
+    let ptr_ty = llvm::r#type::pointer(context, 0);
+    let len_ty = IntegerType::new(context, 32).into();
+    let arr_ty = llvm::r#type::r#struct(context, &[ptr_ty, len_ty, len_ty, len_ty], false);
```
JulianGCalderon: Is `arr_ty` different from `array_type` built on line 79?

src/runtime.rs
```diff
@@ -296,6 +300,110 @@ pub unsafe extern "C" fn cairo_native__dict_get(
...
+    drop(dict_rc);
```
JulianGCalderon: There is no need for calling drop here. It is done automatically.

src/runtime.rs
```diff
@@ -296,6 +300,110 @@ pub unsafe extern "C" fn cairo_native__dict_get(
...
+        capacity: len as u32,
+    }
+}
+
+/// Fills each of the tuples in the array with the corresponding content.
+///
+/// Receives a pointer to the dictionary and moves its entries into the given uninitialized array of
+/// (felt252, T, T) tuples.  The dictionary is iterated and for each element, a tuple is filled with the key
+/// and the value. To fill the tuples, the 'tuple_stride' is used to move the pointer and get the
+/// necessary offset.
```
JulianGCalderon: We don't receive the `tuple_stride` anymore so we can remove the last sentence.

src/libfuncs/squashed_dict.rs
DiegoCivi: Its the same, removed duplicated type declaration in [`e2708d7`](https://github.com/lambdaclass/cairo_native/pull/1446/commits/e2708d7a73d01cc92f09e1cddd16058c1052ae64)

src/runtime.rs
```diff
@@ -296,6 +300,110 @@ pub unsafe extern "C" fn cairo_native__dict_get(
...
+
+        // Save the key and move to the offset of the 'first_value'
+        *key_ptr = *key;
+        let first_val_ptr = key_ptr.byte_add(key_size) as *mut u8;
+        first_val_ptr.write_bytes(0, generic_ty_size);
+
+        // Get the element, move to the offset of the 'last_value' and save the element in that address
+        let element = dict.elements.byte_add(generic_ty_size * elem_index) as *mut u8;
+        let last_val_ptr = first_val_ptr.byte_add(generic_ty_size);
+        std::ptr::swap_nonoverlapping(element, last_val_ptr, generic_ty_size);
```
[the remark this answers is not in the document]
DiegoCivi: But since know we are not dropping the elements of the original dictionary, using `copy_nonoverlapping` wouldn't leave unfreed memory? By swapping we don't have to free anything and drop the dictionary without caring about memory leaks. Maybe I am missing something

JulianGCalderon (commented):
It would be nice have some tests that include snapshots: Create a snapshot, squash the original dictionary, convert to entries, and then drop the original snapshot.

JulianGCalderon (approved):

PR #1465, opened 2025-11-14
Fix BoundedInt Sub libfunc

### LINKED ISSUE #1449: Review bounded_int_sub libfunc
Review the `bounded_int_sub` libfunc, to ensure that all possible inputs are taken into account.

- What types can the libfunc arguments have?
- Is the offset of all bounded int values taken into account?
- Is the value of the return bounded int (if any) computed correctly, according to its offset?

We should:

- Document the signature the signature of the libfunc, including input and output values.
- Document the expected behavior of the libfunc.
- Add  unit/integration tests for edge cases.

DiegoCivi:
# Fix BoundedInt Sub libfunc

The previous implementation of the sub libfunc had some cases in which it failed to return the correct result. We decided to change the algorithm to one that is simpler to understand and is mathematically correct.

We have C = A - B, where all of them are `BoundedInt`. We represent the a `BoundedInt` as: 
- X = Xo + Xd, where 
    - X is a `BoundedInt`.
    - Xo is the lower bound of the `BoundedInt` range.
    - Xd is the offset.

So we want to get `Cd` and we have `Co + Cd = ( Ao + Ad ) - ( Bo + Bd )` -> `Cd = Ao + Ad - Bo - Bd - Co`. 

Closes #1449 

No.

## Checklist

- [ ] Linked to Github Issue.
- [ ] Unit tests added.

### DIFF
--- a/src/libfuncs/bounded_int.rs
+++ b/src/libfuncs/bounded_int.rs
@@ -73,6 +73,22 @@ pub fn build<'ctx, 'this>(
 }
 
 /// Generate MLIR operations for the `bounded_int_add` libfunc.
+///
+/// # Cairo Signature
+///
+/// ```cairo
+/// extern fn bounded_int_add<Lhs, Rhs, impl H: AddHelper<Lhs, Rhs>>(
+///    lhs: Lhs, rhs: Rhs,
+/// ) -> H::Result nopanic;
+/// ```
+///
+/// A number X as a `BoundedInt` is internally represented as an offset Xd from the lower bound Xo.
+/// So X = Xo + Xd.
+///
+/// Since we want to get C = A + B, we can translate this to
+/// Co + Cd = Ao + Ad + Bo + Bd. Where Ao, Bo and Co represent the lower bound
+/// of the ranges in the `BoundedInt` and Ad, Bd and Cd represent the offsets. Since
+/// we also know that Co = Ao + Bo we can simplify the equation to Cd = Ad + Bd.
 #[allow(clippy::too_many_arguments)]
 fn build_add<'ctx, 'this>(
     context: &'ctx Context,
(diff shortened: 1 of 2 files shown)


src/libfuncs/bounded_int.rs
```diff
@@ -1007,6 +988,111 @@ mod test {
...
+    #[test_case("run_test_3", -3, -3, 0, 0, 1)]
+    #[test_case("run_test_4", -6, 3, -9, -9, -3)]
+    #[test_case("run_test_5", -2, -20, 18, 4, 19)]
+    fn test_sub(
+        entry_point: &str,
+        lhs: i32,
+        rhs: i32,
+        result: i32,
+        lower_bound: i32,
+        upper_bound: i32,
```
JulianGCalderon: The lower and upper bound are given by the sierra, so there is no need on checking it, right?

JulianGCalderon (approved):
LGTM!

src/libfuncs/bounded_int.rs
```diff
@@ -276,47 +277,23 @@ fn build_sub<'ctx, 'this>(
...
-    } else {
-        res_value
-    };
-
-    let res_value = if dst_range.offset_bit_width() < compute_range.offset_bit_width() {
+    let compile_time_val =
+        entry.const_int_from_type(context, location, compile_time_val, compute_ty)?;
+    // First we do -> Ad - Bd = intermediate_res
+    let res_value = entry.subi(lhs_value, rhs_value, location)?;
+    // Then we do -> intermediate_res + (Ao - Bo - Co)
```
gabrielbosio: I think it's a bit clearer this way:

```suggestion
    // First we do -> intermediate_res = Ad - Bd
    let res_value = entry.subi(lhs_value, rhs_value, location)?;
    // Then we do -> intermediate_res += (Ao - Bo - Co)
```

JulianGCalderon (approved):

gabrielbosio (approved):

PR #1480, opened 2025-11-27
Fix comparison of `bounded_int_constrain`

DiegoCivi:
# Fix comparison of `bounded_int_constrain`

When checking if the input is lower or greater than the boundary, we were using either `signed less than` or `unsigned less than` depending on the lower bounf of the input. With the fix in #1462 we modified the boundary to represent the offseted value, which will never be negative. So we should always use `unsigned less than` when using `BoundedInt`.

No.

## Checklist

- [ ] Linked to Github Issue.
- [ ] Unit tests added.


FrancoGiachetta (approved):

JulianGCalderon (approved):
LGTM!

PR #1482, opened 2025-11-27
Fix upcast for bounded ints

### LINKED ISSUE #1457: Review upcast libfunc for bounded ints
Review the `upcast` libfunc, to ensure that bounded int values are taken into account

- What types can the libfunc arguments have?
- Is the offset of all bounded int values taken into account?
- Is the value of the return bounded int (if any) computed correctly, according to its offset?

We should:

- Document the signature the signature of the libfunc, including input and output values.
- Document the expected behavior of the libfunc.
- Add unit/integration tests for edge cases.

JulianGCalderon:
# Fix upcast for bounded ints

Closes #1457 

- Improves documentation on bounded ints
- Adds more tests, replacing the old incomplete test.
- Fixes the implementation: we were not offsetting the value when the source value was not offsetted.

No.

## Checklist

- [x] Linked to Github Issue.
- [x] Unit tests added.

### DIFF
--- a/src/libfuncs/cast.rs
+++ b/src/libfuncs/cast.rs
@@ -372,7 +372,7 @@ pub fn build_upcast<'ctx, 'this>(
             };
             let signed_dst_contains_src = signed_dst_range.lower <= src_range.lower
                 && signed_dst_range.upper >= src_range.upper;
-            dst_contains_src | signed_dst_contains_src
+            dst_contains_src || signed_dst_contains_src
         } else {
             dst_contains_src
         };


src/libfuncs/cast.rs
```diff
@@ -405,22 +423,22 @@ pub fn build_upcast<'ctx, 'this>(
         src_value
     };
```
DiegoCivi: Shouldn't we do a `trunci` operation somewhere around here? For example if we have a downcast of a `BoundedInt<0, 10>` to a `BoundedInt<11, 13>`

DiegoCivi: This is upcast, not downcast. I go mixed up, my bad.

src/libfuncs/cast.rs
```diff
@@ -329,7 +329,15 @@ pub fn build_downcast<'ctx, 'this>(
     Ok(())
 }
 
-/// Generate MLIR operations for the `upcast` libfunc.
+/// Builds the `upcast` libfunc, which convert from a source type `T` to a
```
gabrielbosio:
```suggestion
/// Builds the `upcast` libfunc, which converts from a source type `T` to a
```

src/libfuncs/cast.rs
```diff
@@ -350,22 +358,32 @@ pub fn build_upcast<'ctx, 'this>(
...
+            let signed_dst_range = Range {
                 lower: BigInt::from_biguint(Sign::Minus, HALF_PRIME.clone()),
                 upper: BigInt::from_biguint(Sign::Plus, HALF_PRIME.clone()) + BigInt::one(),
             };
-
-            (dst_range.lower <= src_range.lower && dst_range.upper >= src_range.upper)
-                || (alt_range.lower <= src_range.lower && alt_range.upper >= src_range.upper)
+            let signed_dst_contains_src = signed_dst_range.lower <= src_range.lower
+                && signed_dst_range.upper >= src_range.upper;
+            dst_contains_src | signed_dst_contains_src
```
gabrielbosio: Why use the `|` operator instead of `||` in this case?

gabrielbosio (changes requested):
Looks good, left some small comments

FrancoGiachetta (approved):

PR #1484, opened 2025-12-02
Refactor syscalls

FrancoGiachetta:
# Refactor syscalls

Closes #NA

This PR refactors syscalls' execution implementations. 

Key changes: 
- New helper function `execute_syscall` which receives the expected arguments and return types and takes care of calling the syscall and creating the resulting payloads. This way, the amount of code gets drastically reduced.
- Removed `payload_ok` in syscalls where it represents `Ok(())`
- Change the type of `payload_arg` in syscall `send_message_to_l1`. This is because it a `Span<felt252>`, which represents:

    ```rust
    struct {
	    struct {
		    ptr,
		    start,
		    end,
		    capacity
	    }
    }
    ```
    But it was being represented as:
    ```rust
	struct {
		ptr,
		start,
		end,
		capacity
	}
    ```
    Even though this was fine, since the each represent the same pointer, the former is the correct type to use.

No.

## Checklist

- [ ] Linked to Github Issue.
- [ ] Unit tests added.

### DIFF
--- a/.github/ISSUE_TEMPLATE/daily_cache_failure.md
+++ b/.github/ISSUE_TEMPLATE/daily_cache_failure.md
@@ -1,0 +1,11 @@
+---
+title: "CACHE MISS - Daily Workflow Failure"
+labels: bug
+---
+
+Daily workflow failed to retrieve the blocks to execute from the cache. To save the blocks in the cache,
+`cache-vm-daily` workflow should be run manually with the desired blocks. 
+
+Input example: "740000,760000,780000,800000"
+
+- Workflow URL: {{ env.WORKFLOW_URL }}
--- a/.github/workflows/cache-vm-daily.yml
+++ b/.github/workflows/cache-vm-daily.yml
@@ -1,0 +1,165 @@
+name: VM Block Run
+
+on: 
+  workflow_dispatch:
+    inputs:
+      blocks:
+        type: string
+        description: Comma separated block numbers (e.g "10,50,70")
+        required: true
+
+env:
+  RANGE_SIZE: 20
+
+jobs:
+
+  # Gets the blocks to execute and saves them on the cache so it can
+  # be used by the 'daily' workflow
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      blocks: ${{ steps.parse-blocks.outputs.blocks_list }}
+    steps:
+
+      - name: Checkout Native
+        uses: actions/checkout@v4
+        with:
+          path: cairo_native
+
+      - name: Restore blocks list
+        id: cache-restore
+        uses: actions/cache/restore@v4.2.0
+        with:
+          path: ./blocks.txt
+          key: blocks-list
+
+      - name: Delete blocks cache entry
+        if: ${{ steps.cache-restore.outputs.cache-hit }}
+        run: gh cache delete blocks-list
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+
+      - name: Parse blocks
+        id: parse-blocks
+        run: |
+          echo "blocks_list=$(jq 'split(",")' -Rc <(echo ${{ github.event.inputs.blocks }}))" >> $GITHUB_OUTPUT
+          echo "${{ github.event.inputs.blocks }}" > ./blocks.txt
+      
+      - name: Save blocks list
+        uses: actions/cache/save@v4.2.0
+        with:
+          path: ./blocks.txt
+          key: blocks-list
+
+  run:
+    needs: setup
+    runs-on: ubuntu-latest
+    env:
+      LLVM_SYS_191_PREFIX: /usr/lib/llvm-19/
+      MLIR_SYS_190_PREFIX: /usr/lib/llvm-19/
+      TABLEGEN_190_PREFIX: /usr/lib/llvm-19/
+      RPC_ENDPOINT_TESTNET: ${{ secrets.RPC_ENDPOINT_TESTNET }}
+      RPC_ENDPOINT_MAINNET: ${{ secrets.RPC_ENDPOINT_MAINNET }}
+    strategy:
+      max-parallel: 25
+      matrix:
+        block: ${{ fromJson(needs.setup.outputs.blocks) }}
+      fail-fast: false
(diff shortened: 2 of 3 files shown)


src/libfuncs/starknet.rs
```diff
@@ -2980,37 +2097,25 @@ pub fn build_get_class_hash_at<'ctx, 'this>(
     )?;
 
     let payload_ok = {
-        let value = entry.load(
+        let value = block.load(
```
DiegoCivi: I might be wrong, but in many of the syscalls we were moving the pointer with a `.gep()` operation and I think that is not being done anymore. Or is it being done in another place and I am missing it?

src/libfuncs/starknet.rs
```diff
@@ -2980,37 +2097,25 @@ pub fn build_get_class_hash_at<'ctx, 'this>(
...
             location,
             result_ptr,
             llvm::r#type::r#struct(context, &[result_tag_ty, variant_tys[0].0], false),
         )?;
-        entry.extract_value(context, location, value, variant_tys[0].0, 1)?
+        block.extract_value(context, location, value, variant_tys[0].0, 1)?
     };
     let payload_err = {
-        let value = entry.load(
+        let value = block.load(
```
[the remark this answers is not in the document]
FrancoGiachetta: Same answer as above.

src/libfuncs/starknet.rs
```diff
@@ -2943,19 +1992,87 @@ pub fn build_get_class_hash_at<'ctx, 'this>(
...
+            &[remaining_gas, entry.arg(1)?, payload_ok],
+        ],
+        location,
+    )
+}
+
+/// Helper function to call a syscall.
+///
+/// This function receives the necessary arguments needed by the syscall as
+/// well as the return types, and returns the result of its call.
```
gabrielbosio: How about adding the corresponding arguments the docs are referencing? For instance, if I'm not mistaken:

```rust
/// This function receives the necessary arguments `args`
```

This would also apply to the argument/arguments referenced in `return types`

src/libfuncs/starknet.rs
```diff
@@ -2943,19 +1992,87 @@ pub fn build_get_class_hash_at<'ctx, 'this>(
...
+#[allow(clippy::too_many_arguments)]
+fn call_syscall<'ctx, 'this>(
```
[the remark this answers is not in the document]
FrancoGiachetta: Yes, its better. I'll change it

src/libfuncs/starknet.rs
```diff
@@ -729,83 +544,44 @@ pub fn build_storage_write<'ctx, 'this>(
...
+    let remaining_gas = entry.load(context, location, gas_builtin_ptr, gas_ty)?;
+
+    helper.cond_br(
         context,
+        entry,
+        result_tag,
+        [1, 0],
+        [
+            &[remaining_gas, entry.arg(1)?, payload_err],
+            &[remaining_gas, entry.arg(1)?, payload_ok],
```
JulianGCalderon: This branch has 3 values, but the signature only contains 2 values. This is not a problem because `cond_br` ignores the extra argument, but I would add a comment saying that `payload_ok` is garbage and **should not** used.

src/libfuncs/starknet.rs
JulianGCalderon: Why isn't `call_syscall` used here?

JulianGCalderon: Idem here. https://github.com/lambdaclass/cairo_native/pull/1484/files#r2593802052

src/libfuncs/starknet.rs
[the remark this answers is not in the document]
JulianGCalderon: Sometimes this function is used on syscall that don't have a type for the ok variant. We should document what to do in that case.

JulianGCalderon (changes requested):

src/libfuncs/starknet.rs
FrancoGiachetta: Done: [0823ea5](https://github.com/lambdaclass/cairo_native/pull/1484/commits/0823ea52dce2b2d1224dc788070af154692595a1)

src/libfuncs/starknet.rs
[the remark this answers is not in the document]
FrancoGiachetta: Done in: [0823ea5](https://github.com/lambdaclass/cairo_native/pull/1484/commits/0823ea52dce2b2d1224dc788070af154692595a1)

FrancoGiachetta: Don't really know if this is what you meant, but I added more docs here: [0823ea5](https://github.com/lambdaclass/cairo_native/pull/1484/commits/0823ea52dce2b2d1224dc788070af154692595a1)

src/libfuncs/starknet.rs
```diff
@@ -2943,19 +2006,105 @@ pub fn build_get_class_hash_at<'ctx, 'this>(
...
+///
+/// This function receives the necessary arguments needed by the syscall as
+/// well as the return types, and returns the result of its call.
+///
+/// ## Arguments
+///
+/// The `args` vector should **only** hold syscall's arguments in the order
+/// stated by its signature. This function is in charge of creating the return
+/// pointer in which the syscall will store the result. So, the `args` vector
+/// **should not** hold it.
```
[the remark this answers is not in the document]
gabrielbosio: I would clarify it

src/libfuncs/starknet.rs
```diff
@@ -3088,19 +2139,105 @@ pub fn build_get_class_hash_at<'ctx, 'this>(
...
+///
+/// ## Return Types
+///
+/// Cairo Native's syscalls implementations always expect two result branches.
+/// There are some syscalls which don't have a result type for the ok case.
+/// For these cases, the payload ok should be built with a duplicated type from
+/// the error branch. Due to this, it should only be built and **never** be
+/// used.
```
gabrielbosio:
```suggestion
/// Cairo Native's syscalls implementations always expect two result branches.
/// There are some syscalls which don't have a result type for the ok case.
/// For these cases, the payload returned in the ok case should be built with a
/// duplicated type from the error branch. Due to this, the payload should only
/// be built and **not** used in these cases.
```

gabrielbosio (approved):

PR #1510, opened 2025-12-18
Implement get_execution_info_v3 syscall

JulianGCalderon:
# Implement get_execution_info_v3 syscall

Implements the get_execution_info_v3 syscall.

Co-authored-by: @FrancoGiachetta

Yes.

- [x] Created PR in [sequencer](https://github.com/lambdaclass/sequencer)
- [x] Created PR in [starknet-replay](https://github.com/lambdaclass/starknet-replay)   
- [x] Updated the `starknet-blocks.yml` workflow to use these PRs.

### DIFF
--- a/.github/workflows/starknet-blocks.yml
+++ b/.github/workflows/starknet-blocks.yml
@@ -32,7 +32,7 @@ jobs:
         with:
           repository: lambdaclass/starknet-replay
           path: starknet-replay
-          ref: 30020b2865e40be1e6802cb41e189ed44ac2f0c2
+          ref: 71cc48b4bd6acce16f8968760ab98f63c236bc1e
 
       # We need native to use the linux deps ci action
       - name: Checkout Native
@@ -45,7 +45,7 @@ jobs:
           repository: lambdaclass/sequencer
           path: sequencer
           # branch main-v0.14.1
-          ref: 29b213259a5cf4910ea11a05f549dfd5531fc128
+          ref: 6deed4ae852da5b19875c983864e19ebb4b91926
       
       - name: Restore RPC Calls
         id: restore-rpc-calls


examples/erc20.rs
```diff
@@ -85,6 +85,14 @@ impl StarknetSyscallHandler for SyscallHandler {
         })
     }
 
+    fn get_execution_info_v3(
+        &mut self,
+        _remaining_gas: &mut u64,
+    ) -> SyscallResult<cairo_native::starknet::ExecutionInfoV3> {
+        println!("Called `get_execution_info_v3()` from MLIR.");
```
gabrielbosio: Does it have to be implemented in the future?

examples/starknet.rs
```diff
@@ -127,6 +127,14 @@ impl StarknetSyscallHandler for SyscallHandler {
         })
     }
 
+    fn get_execution_info_v3(
+        &mut self,
+        _remaining_gas: &mut u64,
+    ) -> SyscallResult<cairo_native::starknet::ExecutionInfoV3> {
+        println!("Called `get_execution_info_v3()` from MLIR.");
```
gabrielbosio: Does it have to be implemented in the future?

JulianGCalderon: It's an example, so I think that there is no need for implementing. We will need to implement it eventually for the `StubSyscallHandler` though, after PR https://github.com/lambdaclass/cairo_native/pull/1494.

examples/erc20.rs
JulianGCalderon: It's an example, so I think that there is no need for implementing. We will need to implement it eventually for the `StubSyscallHandler` though, after PR https://github.com/lambdaclass/cairo_native/pull/1494.

FrancoGiachetta (approved):

PR #1528, opened 2026-01-07
Add support for precompiling Cairo programs/projects

JulianGCalderon:
# Add support for precompiling Cairo programs/projects

We want to remove the Cairo to Sierra compilation from the testing process. To do so, we compile the Cairo externally, and load it from the tests.

## Cairo Compilation

We will store *all* the Cairo programs in the `test_data` directory, organized by type:

- `programs/`: Cairo programs (e.g. `fibonacci.cairo`)
- `contracts/`: Cairo contracts (e.g. `simple_storage_42.cairo`)
- `tests/`: Cairo program tests (e.g. `corelib`)
- `tests_starknet/`: Cairo contract tests (e.g. `bug_samples`, `cairo_level_tests`)
- `scarb/`: Scarb projects (e.g. `alexandria`)

The `test_data_artifacts` directory contains the compiled artifacts, which are gitignored execept for the Scarb projects:

- `.sierra` files, containing the textual Sierra (good for debugging).
- `.sierra.json` files, containing the raw Sierra.
- `.tests.json` files, containing the test compilation.
- `.contract.json` files, containing the contract class.

With the aid of the `test_utils/compile_test_data.py` script, we will compile *all* the Cairo source code (except for Scarb projects) into the `test_data_artifacts` directory. This will be executed on `make deps`.

With the aid of the `test_utils/compile_scarb_projects.py` script, we will compile *all* the Scarb projects into the `test_data_artifacts/scarb` directory. The resulting files *will* be commited into the repository, and should only be called when updating the Cairo version.

Before, the Corelib was symlinked from cairo2/corelib/ on `make deps`. Now, the corelib is commited directly on the repository. Should be updated by `make pull-external-projects` when upgrading the Cairo version. The `make check-corelib` recipe verifies that its version matches the one specified in the Makefile.

## Test Migration

From Rust's tests, we can load these files to test cairo-native *without* additional Cairo compilation.

- For testing a Sierra program function, we need to call, for example, `load_program("fibonacci")`. This automatically resolves to the compiled program at `test_data_artifacts/programs/fibonacci.sierra.json`.

- For testing a Sierra contract entrypoint, we need to call, for example, `load_contract("simple_storage_42")`. This automatically resolves to the compiled contract at `test_data_artifacts/contracts/simple_storage_42.contract.json`.

- For testing a Scarb project, we need to call, for example, `load_scarb_project("alexandria")`. This automatically resolves to the compiled program at `test_data_artifacts/scarb/alexandria/alexandria.sierra.json`.

I've modified some existing tests to make use of this, as an example.

## Running Cairo Tests

For running, we have a new binary `cairo-native-test-2`. It takes pre-compiled `.tests.json` files as input and runs tests without invoking the Cairo compiler. To run *all* the Cairo tests in the project, we can run `make test-cairo`.

## TODO

- [ ] Remove cairo-native-test and scarb-native-test command, and replace it with `cairo-native-test-2`.
- [ ] Migrate all unit tests to this new flow. This can be done in multiple PRs.
- [ ] Migrate all integration tests to this new flow. This can be done in multiple PRs.
- [ ] Add support for precompiling Sierra to CASM.
 
No.

### DIFF
--- a/.github/workflows/starknet-blocks.yml
+++ b/.github/workflows/starknet-blocks.yml
@@ -32,7 +32,7 @@ jobs:
         with:
           repository: lambdaclass/starknet-replay
           path: starknet-replay
-          ref: 71cc48b4bd6acce16f8968760ab98f63c236bc1e
+          ref: 7ff4d14aa25124dea769826f6011fc5fde91a4ee
 
       # We need native to use the linux deps ci action
       - name: Checkout Native
@@ -45,7 +45,7 @@ jobs:
           repository: lambdaclass/sequencer
           path: sequencer
           # branch main-v0.14.1
-          ref: 6deed4ae852da5b19875c983864e19ebb4b91926
+          ref: a78cfb87923946a6f428506dcfc853f82a1398b6
       
       - name: Restore RPC Calls
         id: restore-rpc-calls
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -485,9 +485,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "cairo-lang-casm"
-version = "2.14.1-dev.1"
+version = "2.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3fa2497d98718d215c1f0eb831adf6992598e5185d6c405eeeaa261869161349"
+checksum = "84e6207b0fd16c0d188e91c81ad4e96bbc6c3cd92e128c832d9fbc87e0a4a11e"
 dependencies = [
  "cairo-lang-utils",
  "indoc",
@@ -499,9 +499,9 @@ dependencies = [
 
 [[package]]
 name = "cairo-lang-compiler"
-version = "2.14.1-dev.1"
+version = "2.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4b84080fc90966f78fd5128391d2480a824806f3657d6ccc9512be63ea08edb1"
+checksum = "e28cc7001d335ea5f8f886e12b1d2ff1cb4d3862d570100ac734ef748eb3f441"
 dependencies = [
  "anyhow",
  "cairo-lang-defs",
@@ -526,9 +526,9 @@ dependencies = [
 
 [[package]]
 name = "cairo-lang-debug"
-version = "2.14.1-dev.1"
+version = "2.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "66d2a69364ddd52b8e5fde8a4f26132e72c3bf94b353f85a2caffa905a4531ed"
+checksum = "3f8fa7f741b25dfe4319c657b32c79ef1fc91c854b45b4d82bb279cba35fbbf5"
 dependencies = [
  "cairo-lang-utils",
  "id-arena",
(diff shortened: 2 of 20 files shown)


src/executor/contract.rs
```diff
@@ -826,81 +826,17 @@ mod tests {
...
-                #[abi(embed_v0)]
-                impl ISimpleStorageImpl of super::ISimpleStorage<ContractState> {
-                    fn get(self: @ContractState, x: felt252) -> (felt252, felt252) {
-                        (x, x * 2)
-                    }
-                }
-            }
-        };
-        program
+        load_contract("simple_storage_dup")
```
FrancoGiachetta: It seems a little weird to me to only receive the name of the test file. I'd prefer using the full/relative path, like it's done in `cairo-vm`.

test_utils/test_cairo.sh
FrancoGiachetta: Same as [comment](https://github.com/lambdaclass/cairo_native/pull/1528/changes#r2683751196)

tests/tests/programs.rs
```diff
@@ -115,24 +116,38 @@ lazy_static! {
...
+        let module_name = "fibonacci".to_string();
+        let runner = SierraCasmRunner::new(
+            program.clone(),
+            Some(Default::default()),
+            Default::default(),
+            None,
+        )
+        .unwrap();
+        (module_name, program, runner)
+    };
```
DiegoCivi: I understand this is intended as an example on how to do it now, but could we instead do this? So all the tests that use `FIB` are already changed:
```rust
pub static ref FIB: (String, Program, SierraCasmRunner) = {
        let versioned_program = load_program("fibonacci");
        let program = versioned_program.into_v1().unwrap().program;
        let module_name = "fibonacci".to_string();
        let runner = SierraCasmRunner::new(
            program.clone(),
            Some(Default::default()),
            Default::default(),
            None,
        )
        .unwrap();
        (module_name, program, runner)
};
```

test_utils/pull_external_projects.py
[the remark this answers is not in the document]
JulianGCalderon: This script does not depend on any compiled target.

test_utils/test_cairo.sh
JulianGCalderon: The target is compiled at the start of the script, so it should never fail.

test_utils/pull_external_projects.py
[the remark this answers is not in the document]
FrancoGiachetta: Right, my bad. I misread.

tests/tests/programs.rs
DiegoCivi: I think it is also being used in [this proptest](https://github.com/lambdaclass/cairo_native/blob/4059216470f3963e6966db77ce6b5b6a5eb1d4f8/tests/tests/programs.rs#L256C1-L279C6). Its only 2 tests but maybe is simpler to just change the static ref

JulianGCalderon: Oh, its true. I missed that usage. In that case it depends on what do we prioritize. By modifying the static we would be making the PR smaller. But in my opinion using a static is harder to follow than just loading the program on each test. I feel like having each test load the program independently is simpler, unless the compilation is taking *too long* and we want to reduce it (by having multiple tests share the same program).

scripts/compile_scarb_projects.py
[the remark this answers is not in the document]
JulianGCalderon: Done

tests/tests/programs.rs
DiegoCivi: I agree!

DiegoCivi (approved):

igaray (approved):

PR #1594, opened 2026-04-19
fixed a bug with array slice index wrapping

TomerStarkware:
# Fix u32 overflow in array_slice bounds check

  build_slice in src/libfuncs/array.rs computed slice_end = slice_start + slice_len in 32-bit arithmetic and then compared it against array_len with Ule. With an attacker-chosen
  slice_len = u32::MAX and slice_start ≥ 1, the sum wraps to a small value and the bounds check passes spuriously. The returned snapshot has end < start, so a subsequent
  array_len (which computes end − start as unsigned) yields 0xFFFFFFFF, and any later array_at / array_pop_* reads off the end of the backing allocation.

  The CASM VM performs the same check in felt/range-check arithmetic and does not wrap, so it correctly panics with "Index out of bounds" — meaning the same Sierra bytecode
  produced different results on Native vs the VM. This was observed in the wild against a problematic class on testnet that the sequencer was executing through Native.

  The fix promotes slice_start, slice_len, and array_len to 64 bits before the add and compare, so the sum can never wrap. The 32-bit arithmetic in the valid branch
  (new_array_start = array_start + slice_start, new_array_end = new_array_start + slice_len) stays as-is because once the 64-bit check has passed both partial sums are provably ≤
   array_end ≤ u32::MAX.

  A regression test was added in tests/tests/arrays.rs (array_slice_u32_overflow) which calls arr.span().slice(1, u32::MAX) on a 2-element array and asserts that Native and the
  VM agree (both panic with the same payload). The test fails on the unfixed build with a gas/output mismatch and passes after the fix.

  A repo-wide audit of arith::addi / .addi(...) sites in src/libfuncs/ confirmed build_slice was the only libfunc with this shape (two user-controlled variable-width integers
  added then compared against a same-width bound).

  No.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1594)

### DIFF
--- a/src/libfuncs/array.rs
+++ b/src/libfuncs/array.rs
@@ -1185,24 +1185,26 @@ pub fn build_slice<'ctx, 'this>(
 
     let slice_start = entry.argument(2)?.into();
     let slice_len = entry.argument(3)?.into();
-    let slice_end = entry.append_op_result(arith::addi(slice_start, slice_len, location))?;
 
-    let slice_lhs_bound = entry.append_op_result(arith::cmpi(
+    // Compute `slice_end = slice_start + slice_len` in 64 bits. In 32-bit
+    // arithmetic the sum wraps and lets an attacker-chosen `slice_len`
+    // sneak past the bound below. Since `slice_len` is unsigned (Sierra
+    // u32) it cannot be negative, so once we've checked
+    // `slice_end <= array_len` we also know `slice_start <= array_len`.
+    let wide_ty = IntegerType::new(context, 64).into();
+    let slice_start_wide = entry.extui(slice_start, wide_ty, location)?;
+    let slice_len_wide = entry.extui(slice_len, wide_ty, location)?;
+    let array_len_wide = entry.extui(array_len, wide_ty, location)?;
+    let slice_end_wide =
+        entry.append_op_result(arith::addi(slice_start_wide, slice_len_wide, location))?;
+
+    let slice_bounds = entry.append_op_result(arith::cmpi(
         context,
         CmpiPredicate::Ule,
-        slice_start,
-        array_len,
+        slice_end_wide,
+        array_len_wide,
         location,
     ))?;
-    let slice_rhs_bound = entry.append_op_result(arith::cmpi(
-        context,
-        CmpiPredicate::Ule,
-        slice_end,
-        array_len,
-        location,
-    ))?;
-    let slice_bounds =
-        entry.append_op_result(arith::andi(slice_lhs_bound, slice_rhs_bound, location))?;
 
     let valid_block = helper.append_block(Block::new(&[]));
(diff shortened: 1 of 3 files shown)


orizi (changes requested):
> @orizi reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1594)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on TomerStarkware).

___
*[`src/libfuncs/array.rs` line 1189 at r1](https://reviewable.io/reviews/starkware-libs/cairo_native/1594#-Oq_OCn14x92Fvc56aIJ:-Oq_OCn2ATgo03ZJN2BL:bci0mce) ([raw file](https://github.com/starkware-libs/cairo_native/blob/5bd65c4284eb8689d81d58a5518e2ae88aa4ef20/src/libfuncs/array.rs#L1189)):*
> ```XML
>     let slice_len = entry.argument(3)?.into();
> 
>     // Compute `slice_end = slice_start + slice_len` in 64 bits. In 32-bit
> ```

additionally doc that the slice len is non-negative - so the test that end is smaller than the arrays end is enough.

orizi (commented):
> @orizi reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1594)* status: all files reviewed, 1 unresolved discussion (waiting on TomerStarkware).

TomerStarkware (commented):
> @TomerStarkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1594)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on orizi).

___
*[`src/libfuncs/array.rs` line 1189 at r1](https://reviewable.io/reviews/starkware-libs/cairo_native/1594#-Oq_OCn14x92Fvc56aIJ:-Oq_QGAm4snOWKpnxNpV:b-896fix) ([raw file](https://github.com/starkware-libs/cairo_native/blob/5bd65c4284eb8689d81d58a5518e2ae88aa4ef20/src/libfuncs/array.rs#L1189)):*
<details><summary><i>Previously, orizi wrote…</i></summary><blockquote>

additionally doc that the slice len is non-negative - so the test that end is smaller than the arrays end is enough.
</blockquote></details>

Done.

orizi (approved):
:lgtm:

> @orizi reviewed 2 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1594#-:-Oq_RCVX8QS3MLWDcxXM:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on TomerStarkware).

PR #1599, opened 2026-04-27
add run_with_libfunc_profile + AotWithProgram variant for ContractExecutor

avi-starkware:
## Summary

Exposes the libfunc-profiling primitives that downstream consumers (e.g. the blockifier in `starkware-libs/sequencer`) currently maintain locally. The profile-collection pattern is callback-driven so the per-call key (tx hash, etc.) stays out of cairo-native.

> [!IMPORTANT]
> Stacked on #1597 and #1598. Merge those first; this PR's diff will narrow once they land.

## Changes

- **`metadata::profiler::Profile`** is now `pub` (was a private type alias).
- **`AotContractExecutor::run_with_libfunc_profile<H, F>`** (gated on `with-libfunc-profiling`, in new file `src/executor/libfunc_profile.rs`) wraps `run`: allocates a unique trace ID, points the executor's `cairo_native__profiler__profile_id` symbol at it, drains the resulting `Profile` after `run` returns, and hands it to a caller-supplied `FnOnce(Profile)`. A `ProfilerGuard` restores the previous trace ID and drops the `LIBFUNC_PROFILE` slot on both the success and unwind paths.
- **`ContractExecutor::AotWithProgram(AotWithProgram { executor, program })`** is a new variant that bundles an AOT executor with the Sierra program it was built from. `From<AotWithProgram>` is provided.
- **`ContractExecutor::run`** dispatches the new variant via `run_with_libfunc_profile` with a no-op profile callback.
- **`ContractExecutor::run_with_profile<H, F>`** is the profile-capturing counterpart of `run`; for non-`AotWithProgram` variants it falls through to `run` (callback never fires).

## Test plan

- [ ] CI passes with default features
- [ ] CI passes with `--features with-libfunc-profiling`
- [ ] CI passes with `--features sierra-emu,with-libfunc-profiling`

## Context

Companion sequencer PR that adopts these APIs and drops the locally-maintained profiling code: starkware-libs/sequencer#13880.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1599)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -122,7 +122,9 @@ with-debug-utils = []
 with-mem-tracing = []
 with-libfunc-profiling = []
 with-segfault-catcher = []
-with-trace-dump = ["dep:sierra-emu"]
+with-trace-dump = ["sierra-emu"]
+# Enables `sierra_emu_bridge` for routing a cairo-native syscall handler into the sierra-emu VM.
+sierra-emu = ["dep:sierra-emu"]
 testing = ["dep:cairo-lang-compiler", "dep:cairo-lang-filesystem"]
 
 [dependencies]
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -11,6 +11,8 @@ pub use self::{
 };
 #[cfg(feature = "sierra-emu")]
 pub use self::contract_executor::EmuContractInfo;
+#[cfg(feature = "with-libfunc-profiling")]
+pub use self::contract_executor::AotWithProgram;
 use crate::{
     arch::{AbiArgument, ValueWithInfoWrapper},
     error::{panic::ToNativeAssertError, Error},
@@ -48,6 +50,8 @@ mod aot;
 mod contract;
 mod contract_executor;
 mod jit;
+#[cfg(feature = "with-libfunc-profiling")]
+mod libfunc_profile;
 
 #[cfg(target_arch = "aarch64")]
 global_asm!(include_str!("arch/aarch64.s"));
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -3,7 +3,14 @@
 //! This module provides methods to execute the programs, either via JIT or compiled ahead
 //! of time. It also provides a cache to avoid recompiling previously compiled programs.
 
-pub use self::{aot::AotNativeExecutor, contract::AotContractExecutor, jit::JitNativeExecutor};
+pub use self::{
+    aot::AotNativeExecutor,
+    contract::AotContractExecutor,
+    contract_executor::ContractExecutor,
+    jit::JitNativeExecutor,
+};
+#[cfg(feature = "sierra-emu")]
+pub use self::contract_executor::EmuContractInfo;
 use crate::{
     arch::{AbiArgument, ValueWithInfoWrapper},
     error::{panic::ToNativeAssertError, Error},
@@ -39,6 +46,7 @@ use std::{alloc::Layout, arch::global_asm, ptr::NonNull};
 
 mod aot;
 mod contract;
+mod contract_executor;
 mod jit;
 
 #[cfg(target_arch = "aarch64")]
--- a/src/executor/contract_executor.rs
+++ b/src/executor/contract_executor.rs
@@ -3,21 +3,26 @@
 //!
 //! The `Emu` variant is gated on the `sierra-emu` feature and uses
 //! [`crate::sierra_emu_bridge::SierraEmuSyscallBridge`] to thread a cairo-native syscall
-//! handler through the sierra-emu VM.
+//! handler through the sierra-emu VM. The `AotWithProgram` variant is gated on
+//! `with-libfunc-profiling` and bundles an [`AotContractExecutor`] with the Sierra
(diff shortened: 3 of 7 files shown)


orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1599)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on avi-starkware, TomerStarkware, and Yoni-Starkware).

___
*[`src/sierra_emu_bridge.rs` line 23 at r1](https://reviewable.io/reviews/starkware-libs/cairo_native/1599#-OrOAEwX5BKs2reRDeqf:-OrOAEwX5BKs2reRDeqg:bjnosqt) ([raw file](https://github.com/starkware-libs/cairo_native/blob/a7f3eb800c48aa426b322375d35bd73c89332361/src/sierra_emu_bridge.rs#L23)):*
> ```XML
> pub struct SierraEmuSyscallBridge<'a, H>(pub &'a mut H);
> 
> impl<H: native::StarknetSyscallHandler> emu::StarknetSyscallHandler
> ```

sounds like this trait should be in a base crate that both native and emu depends on - instead of a full fuplication.



_Code quote:_
```XML
StarknetSyscallHandler
```

avi-starkware:
*[`src/sierra_emu_bridge.rs` line 23 at r1](https://reviewable.io/reviews/starkware-libs/cairo_native/1599#-OrOAEwX5BKs2reRDeqf:-OsGgIW85tpxBBjQvyDe:b-srxajf) ([raw file](https://github.com/starkware-libs/cairo_native/blob/a7f3eb800c48aa426b322375d35bd73c89332361/src/sierra_emu_bridge.rs#L23)):*
<details><summary><i>Previously, orizi wrote…</i></summary><blockquote>

sounds like this trait should be in a base crate that both native and emu depends on - instead of a full fuplication.
</blockquote></details>

The main issue is that sierra-emu has its own versions of basic types (SyscallResult, ExecutionInfo, ExecutionInfoV2, U256), and uses Vec instead of slices for arrays of Felts and u64. Moreover, the StarknetSyscallHandler trait is different for cairo_native and emu in several small ways (e.g., cairo-native has get_exuction_info_v3 while sierra-emu doesn't; sierra-emu has cheatcode while cairo-native doesn't; the signature of sha256_process_block is different).

I think the way to cleanly share code between the two would be to create a base types crates (similar to starknet-api) and have both crates depend on it. It is a much bigger refactor than what I am proposing here though.

orizi (changes requested):
> @orizi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1599)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on avi-starkware, TomerStarkware, and Yoni-Starkware).

___
*[`src/sierra_emu_bridge.rs` line 23 at r1](https://reviewable.io/reviews/starkware-libs/cairo_native/1599#-OrOAEwX5BKs2reRDeqf:-OsGoZS8Cc0DG8H9oQAr:b60eeej) ([raw file](https://github.com/starkware-libs/cairo_native/blob/a7f3eb800c48aa426b322375d35bd73c89332361/src/sierra_emu_bridge.rs#L23)):*
<details><summary><i>Previously, avi-starkware (Avi Cohen) wrote…</i></summary><blockquote>

The main issue is that sierra-emu has its own versions of basic types (SyscallResult, ExecutionInfo, ExecutionInfoV2, U256), and uses Vec instead of slices for arrays of Felts and u64. Moreover, the StarknetSyscallHandler trait is different for cairo_native and emu in several small ways (e.g., cairo-native has get_exuction_info_v3 while sierra-emu doesn't; sierra-emu has cheatcode while cairo-native doesn't; the signature of sha256_process_block is different).

I think the way to cleanly share code between the two would be to create a base types crates (similar to starknet-api) and have both crates depend on it. It is a much bigger refactor than what I am proposing here though.
</blockquote></details>

i do not agree it is a bigger refactor. this refactor is just made obsolete by it.

it would require:
1. PR to make the traits exactly the same structure-wise.
2. PR to create a new crate.
3. PRs to use that new crate.

avi-starkware:
*[`src/sierra_emu_bridge.rs` line 23 at r1](https://reviewable.io/reviews/starkware-libs/cairo_native/1599#-OrOAEwX5BKs2reRDeqf:-OsGopIbEySIypgbGrtV:b-6ml0be) ([raw file](https://github.com/starkware-libs/cairo_native/blob/a7f3eb800c48aa426b322375d35bd73c89332361/src/sierra_emu_bridge.rs#L23)):*
<details><summary><i>Previously, orizi wrote…</i></summary><blockquote>

i do not agree it is a bigger refactor. this refactor is just made obsolete by it.

it would require:
1. PR to make the traits exactly the same structure-wise.
2. PR to create a new crate.
3. PRs to use that new crate.
</blockquote></details>

Changing the traits might lead to many other small changes in the crates' implementation, as opposed to just using the crates as they are already implemented. That is why I said it is a bigger refactor.

Anyway, you are right that it is the better way. I will give it a try.

PR #1646, opened 2026-06-18
fix: size dict elements buffer to HashMap capacity, not len

TomerStarkware:
## Summary

When materializing a `Felt252Dict` value (`Value::Felt252Dict` → native), the elements buffer was arena-allocated with room for `map.len()` slots. The runtime's `dict_get` only grows the elements buffer once the backing `HashMap` is full, so a runtime insert of a new key could write one slot past the end of the buffer and corrupt the arena.

This reorders the allocation so the `FeltDict` struct is created and `reserve`d first, then sizes the elements buffer to the HashMap's actual `capacity()` (which `reserve` usually over-allocates). That guarantees there is always room for the runtime to fill every slot the HashMap can hold.

## Changes

- `src/values.rs`: allocate/register the `FeltDict` before the elements buffer, and size the buffer to `dict.mappings.capacity()` instead of `map.len()`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1646)

### DIFF
--- a/src/values.rs
+++ b/src/values.rs
@@ -370,17 +370,6 @@ impl Value {
                         let elem_ty = registry.get_type(&info.ty)?;
                         let elem_layout = elem_ty.layout(registry)?;
 
-                        // Arena-allocate the elements buffer.
-                        let elements = if map.is_empty() {
-                            null_mut()
-                        } else {
-                            crate::runtime::cairo_native__arena_alloc(
-                                (elem_layout.pad_to_align().size() * map.len()) as u64,
-                                elem_layout.align() as u64,
-                            )
-                            .cast()
-                        };
-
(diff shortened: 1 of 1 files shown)


orizi (approved):
:lgtm:

> @orizi reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1646#-:-OvOqecbAL0_IpSlMpZf:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on TomerStarkware).

PR #1650, opened 2026-07-19
fix(values): un-box memory-allocated enum payload in Value::to_ptr

TomerStarkware:
## What

The `Enum` arm of `Value::to_ptr` (`src/values.rs`) copied the payload straight from the pointer returned by the payload's own `to_ptr`. When the payload is a **memory-allocated** type — e.g. a nested `>=2`-variant enum, which `is_memory_allocated` marks true — that pointer is a *wrapper* (a pointer to the data pointer). The copy therefore read the 8-byte wrapper pointer plus adjacent bump-arena garbage into the parent variant slot, instead of the actual payload data.

Because a 2-variant enum accepts any tag bit pattern, the corrupted value never trips an "invalid tag" assert: execution completes and returns a wrong result with **no crash and no signal** — a silent correctness bug.

The sibling `Struct` arm a few lines above already handles this by un-boxing memory-allocated members; this change makes the `Enum` arm do the same.

## Fix

```rust
let payload_type_id = &info.variants[*tag];
let payload_ty = registry.get_type(payload_type_id)?;
let payload = value.to_ptr(arena, registry, payload_type_id)?;
// Undo the wrapper pointer added when the payload is memory
// allocated (e.g. a nested >=2-variant enum).
let payload = if payload_ty.is_memory_allocated(registry)? {
    *payload.cast::<NonNull<()>>().as_ref()
} else {
    payload
};
```

## Reachability / scope

Only reachable through the generic `JitNativeExecutor::invoke_dynamic(&[Value], …)` API (used by `cairo-native-run` / `cairo-native-test` and library consumers passing structured `Value` arguments). The Starknet **contract** path (`executor/contract.rs`) marshals arguments and results as raw felts and never calls `Value::to_ptr` / `Value::from_ptr`, so sequencer/contract execution is unaffected.

## Test

Adds `nested_enum_argument_matches_vm` (`tests/tests/enums.rs`) with a nested 2-variant-enum program (`test_data/programs/nested_enum_arg.cairo`), comparing native vs VM across all four `(outer_tag, inner_tag)` combinations.

- **Before the fix:** fails — for `Outer::X(Inner::A(0x1234))` the VM returns `4660` while native returns garbage (`859618275080168…`).
- **After the fix:** passes; full enum/result suite green (29/29).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1650)

### DIFF
--- a/src/values.rs
+++ b/src/values.rs
@@ -327,7 +327,17 @@ impl Value {
                         native_assert!(*tag < info.variants.len(), "Variant index out of range.");
 
                         let payload_type_id = &info.variants[*tag];
+                        let payload_ty = registry.get_type(payload_type_id)?;
                         let payload = value.to_ptr(arena, registry, payload_type_id)?;
+
+                        // Undo the wrapper pointer added when the payload is memory
+                        // allocated (e.g. a nested >=2-variant enum), so that the copy
+                        // below reads the payload data rather than the wrapper pointer.
+                        let payload = if payload_ty.is_memory_allocated(registry)? {
+                            *payload.cast::<NonNull<()>>().as_ref()
+                        } else {
+                            payload
+                        };
 
                         let (layout, tag_layout, variant_layouts) =
                             crate::types::r#enum::get_layout_for_variants(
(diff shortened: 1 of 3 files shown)


orizi (approved):
:lgtm:

> @orizi reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/cairo_native/1650#-:-Oxuo0Wa-2DZUhEc5FBv:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on TomerStarkware).

